Issue #8, opened 2016-09-29
bug


longinhit:
When the samp_rate is greater than 50, create_attrs_for_span () calculated is_sampled all true

PR #9, opened 2016-09-29
Update sampling algorithm to use random.random()

mjbryant:
Addresses https://github.com/Yelp/py_zipkin/issues/8

### DIFF
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import functools
+import random
 import time
 from collections import namedtuple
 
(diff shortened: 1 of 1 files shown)


kaisen:
🚢

prat0318:
the rationale behind the previous logic was to make the selection deterministic. So, if we wanted to not change sample ratio we can chose a trace id such that it would result in selection. 

If trace_ids are random, then it would lead to correct randomness in selections as well, but if they are not, then it might lead to what is seen in #8.

kaisen:
@prat0318 heh, @mbryant and I talked about that particular reason during lunch :) It's an useful way of ensuring a particular trace_id is sampled (as long as the sampling rate stays constant) but there are better ways to force a trace now.

prat0318:
++

PR #10, opened 2016-09-30
Handle when zipkin_attrs are passed in and sampled correctly

mjbryant:
When sampled zipkin_attrs are passed into `zipkin_span` along with a sample_rate param, new zipkin_attrs are generated, even when they shouldn't be. This fixes that. 

I also renamed the 'is_root' to 'perform_logging' and moved the logic around a bit. I can't recall if we've discussed this before, but I still find the 'is_root' variable name to be confusing. I'm open to changing that part back if anyone feels strongly.

Finally, added a regression test for the correct zipkin_attrs + sampling logic.

### DIFF
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -116,8 +116,12 @@ class zipkin_span(object):
         :type binary_annotations: dict of str -> str
         :param port: The port number of the service. Defaults to 0.
         :type port: int
-        :param sample_rate: Custom sampling rate (between 100.0 and 0.0) if
-                            this is the root of the trace
+        :param sample_rate: Rate at which to sample; 0.0 - 100.0. If passed-in
+            zipkin_attrs have is_sampled=False and the sample_rate param is > 0,
+            a new span will be generated at this rate. This means that if you
+            propagate sampling decisions to downstream services, but still have
+            sample_rate > 0 in those services, the actual rate of generated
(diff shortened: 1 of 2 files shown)


py_zipkin/zipkin.py
```diff
@@ -157,21 +157,19 @@ def __enter__(self):
...
+        # If zipkin_attrs are passed in or this span is doing its own sampling,
+        # it will need to actually log spans at __exit__.
+        self.perform_logging = self.zipkin_attrs or self.sample_rate is not None
 
-        if self.zipkin_attrs:
-            self.is_root = True
         if self.sample_rate is not None:
-            self.is_root = True
-
+            # This clause allows for sampling this service independently
```
kaisen: just to clarify for my own understanding, this means that sample_rate is the sampling rate of all requests that are not already sampled by a parent caller? so the actual tracing rate of the service is potentially higher than the specified sample_rate.

kaisen: if so i think it'll be worth mentioning this in the docstring for the sample_rate param

mjbryant: Yes, this is true, but it was true before this PR. I will update the docs.

kaisen: oh yeah 😣  thanks for adding documentation that i should have added! 😅

kaisen:
:shipit:

Issue #11, opened 2016-10-11
change README to suggest http vs kafka


adriancole:
While Kafka is a great transport, it isn't the best one to start with. How about if we change the example in the README to use http?

We can of course say in prod you'd likely use Kafka, but suggesting http in the README will reduce time spent helping users get started in gitter.

cc @luukhoai who is trying to get started

luukhoai:
Hi, please tell me how to use http instead of Kafka.
Thank you so much

luukhoai:
Hi, 
There is one problem when I use Kafka 
**getaddrinfo failed for http://localhost:9411, exception was [Errno -2] Name or service not known. Is your advertised.host.name correct and resolvable?**

Please tell me, how to figure out it ? Thank you so much

adriancole:
http://localhost:9411 is a zipkin server url (http).

Kafka is a different kind of server. Please wait for instructions on how to
use http, or use the scribe instructions I sent you earlier.

Issue #6, opened 2016-09-21
Support sending spans in batches


mjbryant:
The Zipkin Kafka collector supports receiving batches of spans instead of single spans after https://github.com/openzipkin/zipkin/pull/995. However, having a py_zipkin user implement batching solely in the transport_handler seems like it'd be too difficult on their part. 

Maybe we could support having `ZipkinLoggingContext` take a batch_size param that defaults to something large but sane (20?), then in [log_spans](https://github.com/Yelp/py_zipkin/blob/8c03eac8b0d49fa64d644bef62edbfc044078bb1/py_zipkin/logging_helper.py#L75) we could call transport_handler with a list of thrift-encoded spans as opposed to a single span at a time.

@kaisen, thoughts?

dmitry-prokopchenkov:
@kaisen, @bplotnick What about this? AFAIK Zipkin serer can process batch spans. At least we might modify  [log_spans](https://github.com/Yelp/py_zipkin/blob/8c03eac8b0d49fa64d644bef62edbfc044078bb1/py_zipkin/logging_helper.py#L75) to send root span and self.log_handler.client_spans together. What do you think? I could give a try on this.

dmitry-prokopchenkov:
@kaisen, @bplotnick Guys, your thoughts on this?

bplotnick:
@dmitry-prokopchenkov Go for it!

You could make `log_span` just write to a queue and flush when it hits a configurable batch size (with a sane default). Then at the end of `log_spans`, you flush the queue to write the remaining spans.

dmitry-prokopchenkov:
@bplotnick ok, thanks!

dmitry-prokopchenkov:
@bplotnick just a reminder, this can be closed

PR #12, opened 2016-10-11
Add HTTP transport example first in the README.

mjbryant:


mjbryant:
Addresses #11

mjbryant:
@adriancole want to take a look? How I encode the span in a list is kinda hacky, but the collector wasn't accepting non-list values.

README.md
```diff
@@ -109,8 +109,30 @@ Transport
...
+argument - the thrift-encoded bytes.
+
+The simplest way to get spans to the collector is via HTTP POST. Here's an
+example of a simple HTTP transport using the `requests` library. This assumes
+your Zipkin collector is running at localhost:9411.
+
+```python
+import requests
+
+def http_transport(message):
```
adriancole: Nit since the input is a single span, I'd probably name it encoded_span or just span. In zipkin we call the payload sent message

adriancole:
Thanks! This taught me that the input is a single span (vs a "bundle" aka
list of them). Somehow I guessed the input was a list already, but yeah I
can see how someone can write a async/batch handler.

README.md
luukhoai: Hi, I checked this code above. However, I didn't see where to input service_name & span_name.
It also not working. Thanks

luukhoai: screenshot from 2016-10-12 08 15 10

Hi, I checked. but it raise new error: py_zipkin.exception.ZipkinError: Root spans require a transport handler to be given. 
Please check, thanks so much :D

kaisen: the transport handler is supposed to be a function, so try passing in http_transport as the parameter value instead.

transport_handler=http_transport,

luukhoai: localhost:9411 also not working too. Thanks

mjbryant:
@adriancole yeah, our transport handler just takes a single span. The user can implement batching. I still want to implement #6 to change this to a list of spans, though. Seems strictly more flexible and should make implementing a batched transport handler a bit easier.

README.md
luukhoai: It's working. THank you so much :D

adriancole:
I made an example out of this. if you don't mind, please take a look! https://github.com/openzipkin/pyramid_zipkin-example/pull/1

Issue #13, opened 2016-10-13
how do we idiomatically send cs, cr annotations


adriancole:
hey there. In making this, I noticed the resulting span for an outgoing request has only server annotations ("sr", "ss"). https://github.com/openzipkin/pyramid_zipkin-example/pull/1

do you have ideas on making client instrumentation as an alternative to this pattern?

``` python
    headers = {}
    headers.update(create_http_headers_for_new_span())
    backend_response = requests.get(
        url='http://localhost:9000/api',
        headers=headers,
    )
```

Ex some interceptor for requests that starts a span, logs "cs" in that span, makes the outgoing request, logs "cr" in the same span once received, potentially adding an error annotation on transit error.

adriancole:
one idea is to make a generic filter which adds annotations within a span for a scoped operation..

```
before: Annotation <- added at the beginning of a task
after: Annotation <- added unconditionally after a task
after_failure: Annotation + ": $error.class: $error.message" <- additionally added when the task failed
```

ex. I could add do something like (forgive the quasi code)
```
start_new_span
  filter("cs", "cr", "Client Send Error: $error.class: $error.message")
    httpclient.get
        filter("ws", "wr", "Wire Send Error: $error.class: $error.message")
          http_transport.send
```

In this case, the span above would include in normal case: ["cs", "ws", "wr", "cr"]

This is the design finagle uses, and elaborated here
https://github.com/openzipkin/openzipkin.github.io/issues/52

mjbryant:
I'm not totally sure what you're asking/suggesting here. I think there might be a few separate things: how (and whether) pyramid-zipkin logs (1) client spans, (2) error annotations, and (3) wire events.

Re (1): this library actually has no support for logging client spans directly. The `zipkin_span` decorator/context-manager, when used inside an existing active span context, will log both server and client annotations. This is probably bad - we use this internally to give more structure to our traces (e.g. with this pattern a function call can easily be its own span). But `zipkin_span`, when used not in an existing active trace, will log only server annotations. This is because we extracted this code from pyramid-zipkin, which was only ever meant to add server-side instrumentation.

In the example using `requests`, we could have theoretically used an instrumented version of this library, but that'd require some additional code.

[...]

PR #14, opened 2016-10-17
add annotation filters, zipkin_client_span, and zipkin_server_span

bplotnick:
This is a pretty rough attempt at adding support for client-only or server-only spans. I tried to base it loosely on how finagle does filtering (referenced in #13) - that is, have standard span subclasses that filter the annotations to collect. The big difference being that we still allow both client and server annotations in the span (the main case being method annotations or any annotations that are not inter-process)

It doesn't address error annotations or wire events. It also doesn't address the root span, which we currently still enforce to be `ss`/`sr` only.

### DIFF
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -163,8 +163,8 @@ class zipkin_span(object):
             # get a list of all of the mapped annotations
             self.annotation_filter = {
                 annotation
-                for include_name, annotation_list in STANDARD_ANNOTATIONS.items()
-                for annotation in annotation_list
+                for include_name, annotation_set in STANDARD_ANNOTATIONS.items()
+                for annotation in annotation_set
                 if include_name in include
             }
 
@@ -323,14 +323,44 @@ class zipkin_span(object):
         self.logging_context.binary_annotations_dict.update(extra_annotations)
 
 
+def _validate_args(kwargs):
+    if 'include' in kwargs:
+        raise ValueError(
+            '"include" is not valid in this context.'
+            'You probably want to use zipkin_span()'
+        )
+
+
 class zipkin_client_span(zipkin_span):
+    """Logs a client-side zipkin span.
+
+    Subclass of :class:`zipkin_span` using only annotations relevant to clients
+    """
+
     def __init__(self, *args, **kwargs):
+        """Logs a zipkin span with client annotations.
+
+        See :class:`zipkin_span` for arguments
+        """
+        _validate_args(kwargs)
+
         kwargs['include'] = ('client',)
         super(zipkin_client_span, self).__init__(*args, **kwargs)
 
 
 class zipkin_server_span(zipkin_span):
+    """Logs a server-side zipkin span.
+
+    Subclass of :class:`zipkin_span` using only annotations relevant to servers
+    """
+
     def __init__(self, *args, **kwargs):
+        """Logs a zipkin span with server annotations.
+
+        See :class:`zipkin_span` for arguments
+        """
+        _validate_args(kwargs)
+
         kwargs['include'] = ('server',)
         super(zipkin_server_span, self).__init__(*args, **kwargs)
 
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -30,6 +30,12 @@ ZipkinAttrs = namedtuple(
 )
 
 
+STANDARD_ANNOTATIONS = {
+    'client': {'cs', 'cr'},
+    'server': {'ss', 'sr'},
+}
+
+
 class zipkin_span(object):
     """Context manager/decorator for all of your zipkin tracing needs.
 
@@ -97,6 +103,7 @@ class zipkin_span(object):
         binary_annotations=None,
         port=0,
         sample_rate=None,
+        include=('client', 'server'),
     ):
         """Logs a zipkin span. If this is the root span, then a zipkin
         trace is started as well.
@@ -123,6 +130,10 @@ class zipkin_span(object):
             sample_rate > 0 in those services, the actual rate of generated
             spans for those services will be > sampling_rate.
         :type sample_rate: float
+        :param include: which annotations to include
+            can be one of {'client', 'server'}
+            corresponding to ('cs', 'cr') and ('ss', 'sr') respectively
+        :type include: iterable
         """
         self.service_name = service_name
         self.span_name = span_name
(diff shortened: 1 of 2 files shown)


py_zipkin/zipkin.py
```diff
@@ -143,6 +154,20 @@ def __init__(
...
+        if not set(include).issubset(set(STANDARD_ANNOTATIONS.keys())):
+            raise ZipkinError(
+                'Only %s are supported as annotations' %
+                STANDARD_ANNOTATIONS.keys()
+            )
+        else:
+            # get a list of all of the mapped annotations
+            self.annotation_filter = {
+                annotation
+                for include_name, annotation_list in STANDARD_ANNOTATIONS.items()
```
mjbryant: Nit: `annotation_list` isn't a list, it's a set.

py_zipkin/zipkin.py
```diff
@@ -143,6 +154,20 @@ def __init__(
...
             raise ZipkinError('Sample rate must be between 0.0 and 100.0')
 
+        if not set(include).issubset(set(STANDARD_ANNOTATIONS.keys())):
+            raise ZipkinError(
+                'Only %s are supported as annotations' %
+                STANDARD_ANNOTATIONS.keys()
+            )
+        else:
+            # get a list of all of the mapped annotations
+            self.annotation_filter = {
```
mjbryant: I kinda don't like this implementation on principle, because it's O(number of types of standard annotations) rather than O(number of included annotation types). But this is probably a silly reason to dislike it, since I really doubt we'll ever end up with more than a tiny amount of standard annotations.

py_zipkin/zipkin.py
```diff
@@ -292,6 +323,18 @@ def update_binary_annotations_for_root_span(self, extra_annotations):
...
 
+class zipkin_client_span(zipkin_span):
+    def __init__(self, *args, **kwargs):
+        kwargs['include'] = ('client',)
+        super(zipkin_client_span, self).__init__(*args, **kwargs)
+
+
+class zipkin_server_span(zipkin_span):
+    def __init__(self, *args, **kwargs):
+        kwargs['include'] = ('server',)
```
mjbryant: Could you check for 'include' in kwargs and error if it's there? Does that sound like a good idea?

py_zipkin/zipkin.py
```diff
@@ -292,6 +323,18 @@ def update_binary_annotations_for_root_span(self, extra_annotations):
         self.logging_context.binary_annotations_dict.update(extra_annotations)
 
 
+class zipkin_client_span(zipkin_span):
+    def __init__(self, *args, **kwargs):
+        kwargs['include'] = ('client',)
```
mjbryant: Could you check for 'include' in kwargs and error if it's there? Does that sound like a good idea?

py_zipkin/zipkin.py
```diff
@@ -292,6 +323,18 @@ def update_binary_annotations_for_root_span(self, extra_annotations):
         self.logging_context.binary_annotations_dict.update(extra_annotations)
 
 
+class zipkin_client_span(zipkin_span):
+    def __init__(self, *args, **kwargs):
```
mjbryant: Docstring?

py_zipkin/zipkin.py
```diff
@@ -292,6 +323,18 @@ def update_binary_annotations_for_root_span(self, extra_annotations):
...
+        kwargs['include'] = ('client',)
+        super(zipkin_client_span, self).__init__(*args, **kwargs)
+
+
+class zipkin_server_span(zipkin_span):
+    def __init__(self, *args, **kwargs):
```
mjbryant: Docstring?

tests/zipkin_test.py
```diff
@@ -377,6 +389,74 @@ def test_span_context(
...
 
+@mock.patch('py_zipkin.thread_local._thread_local', autospec=True)
+@mock.patch('py_zipkin.zipkin.generate_random_64bit_string', autospec=True)
+@mock.patch('py_zipkin.zipkin.zipkin_logger', autospec=True)
+def test_client_span_context(
+    zipkin_logger_mock,
+    generate_string_mock,
+    thread_local_mock,
+):
+    zipkin_attrs = ZipkinAttrs(
```
mjbryant: Could you pull the first many lines of this test case into a helper function or a fixture? It's so very DRY.

py_zipkin/zipkin.py
```diff
@@ -323,14 +323,44 @@ def update_binary_annotations_for_root_span(self, extra_annotations):
         self.logging_context.binary_annotations_dict.update(extra_annotations)
 
 
+def _validate_args(kwargs):
+    if 'include' in kwargs:
+        raise ValueError(
+            '"include" is not valid in this context.'
```
mjbryant: This'll print weird 'cause you're missing a space after the period.

py_zipkin/zipkin.py
mjbryant: Agreed. I think it's fine as is. You could probably write it as a map, but the simplest thing I can come up with is:

```python
self.annotation_filter = set()
map(lambda include_name: self.annotation_filter.update(STANDARD_ANNOTATIONS[include_name]), include)
```
which is considerably worse. I mean, you could also just do:
```python
self.annotations = set()
for include_name in include:
    self.annotations.update(STANDARD_ANNOTATIONS[include_name])
```

This has the nice property of not allowing `include` values that aren't in `STANDARD_ANNOTATIONS`, and is way more readable. Doesn't feel as pythonic, though.

bplotnick: I like this much better!

py_zipkin/zipkin.py
```diff
@@ -143,6 +154,17 @@ def __init__(
         if self.sample_rate is not None and not (0.0 <= self.sample_rate <= 100.0):
             raise ZipkinError('Sample rate must be between 0.0 and 100.0')
 
+        if not set(include).issubset(set(STANDARD_ANNOTATIONS.keys())):
```
mjbryant: Sorry for the late comment. It'd be nice if the `set(STANDARD_ANNOTATIONS.keys())` happened out of the context of this function. Something like `ANNOTATION_KEYS = fronzenset(STANDARD_ANNOTATIONS.keys())` in the global scope. Sound reasonable? The less work we can do at runtime the better, IMO.

py_zipkin/zipkin.py
```diff
@@ -143,6 +154,17 @@ def __init__(
...
+            raise ZipkinError(
+                'Only %s are supported as annotations' %
+                STANDARD_ANNOTATIONS.keys()
```
mjbryant: If you do the above comment on moving the keys set generation to global scope you can replace this.

py_zipkin/zipkin.py
bplotnick: I considered this, but I thought it to be premature optimization.

```
$ python -m timeit -s 'x={"a":{"a1","a2","a3"},"b":{"b1","b2","b3"},"c":{"c1","c2","c3"},"d":{"d1","d2","d3"},"e":{"e1","e2","e3"}}' 'y=set(x.keys())'
1000000 loops, best of 3: 0.476 usec per loop
```

But since I don't think it hurts readability to make the change, i'll go ahead and make it.

mjbryant:
<3. :shipit:

kaisen:
🚢

Issue #15, opened 2016-10-19
override autocalc of host in thrift.create_endpoint


lorinbeer:
currently host is determined with socket.gethostbyname

change create_endpoint to accept a 'host' param to override the default behaviour

PR #16, opened 2016-10-19
[#15] add host param for thrift create_endpoint

lorinbeer:
overrides behaviour of determining host ip
backwards compatible with previous interface
unit test for host parameter in create_endpoint


kaisen:
lg2m. i recommend a 0.x version bump since this changes the interface of a function.

lorinbeer:
I'll merge it in, second on the 0.x version bump as it adds functionality but is backwards compatible

PR #17, opened 2016-10-20
start and stop span manual

GuoJing:
Context manager is good but not easy for some framework. This should be easy for some Python framework like flask:

```python
#some middleware
@bp.before_request
def before():
    span.start()

@bp.after_request
def before():
    span.stop()
```

### DIFF
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -174,6 +174,9 @@ class zipkin_span(object):
         return decorated
 
     def __enter__(self):
+        return self.start()
+
+    def start(self):
         """Enter the new span context. All annotations logged inside this
         context will be attributed to this span. All new spans generated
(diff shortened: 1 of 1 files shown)


mjbryant:
You can do this right now like:

```python
span = zipkin_span('service')
span.__enter__()
span.__exit__(None, None, None)
```

But this _is_ cleaner. I'm okay with adding this nicer syntax. @bplotnick, @kaisen: thoughts?

bplotnick:
👍 

This seems much cleaner than calling the enter/exit functions

kaisen:
no objections here. looks like nice syntactic sugar.

mjbryant:
This is available on PyPI as v0.4.0.

GuoJing:
Thanks guys!

PR #18, opened 2016-10-24
Allow overriding the span_id when generating ZipkinAttrs

kaisen:
There are situations where logging multiple spans with the same span_id is useful (such as when you want to log multiple spans to the same parent span in different parts of the code base). You could explicitly construct the correct ZipkinAttrs object each time but it's a bit easier to have create_attrs_for_span take a span_id param.


mjbryant:
👍

PR #21, opened 2016-11-03
Decorators now create new zipkin_span instances instead of using itself

### LINKED ISSUE #20: In rare cases, a zipkin_span context that isn't the root span can try to log using transport_handler
```
...
File "/venv/lib/python2.7/site-packages/py_zipkin/zipkin.py", line 173, in decorated
  return f(*args, **kwargs)
File "/venv/lib/python2.7/site-packages/py_zipkin/zipkin.py", line 268, in __exit__
  self.stop(_exc_type, _exc_value, _exc_traceback)
File "/venv/lib/python2.7/site-packages/py_zipkin/zipkin.py", line 286, in stop
  self.logging_context.__exit__(_exc_type, _exc_value, _exc_traceback)
File "/venv/lib/python2.7/site-packages/py_zipkin/logging_helper.py", line 72, in __exit__
  self.log_spans()
File "/venv/lib/python2.7/site-packages/py_zipkin/logging_helper.py", line 162, in log_spans
  transport_handler=self.transport_handler,
File "/venv/lib/python2.7/site-packages/py_zipkin/logging_helper.py", line 281, in log_span
  transport_handler(message)
TypeError: 'NoneType' object is not callable
```

We've encountered rare situations where a block of code is wrapped with a zipkin_span context within the context of a root span, but when the inner zipkin_span context exits, it somehow thinks it's a root span and tries to log stuff using the transport_handler. But since transport_handler is null, this raises an error.

I haven't been able to duplicate this yet, but a bandaid fix is to be defensive about transport_handler being null when attempting to log spans.

ericchu94:
This fixes #20
I believe the work around implemented in #19 is no longer needed

### DIFF
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -145,6 +145,7 @@ class zipkin_span(object):
         self.port = port
         self.logging_context = None
         self.sample_rate = sample_rate
+        self.include = include
 
         # Validation checks
         if self.zipkin_attrs or self.sample_rate is not None:
@@ -169,7 +170,17 @@ class zipkin_span(object):
     def __call__(self, f):
         @functools.wraps(f)
         def decorated(*args, **kwargs):
-            with self:
+            with zipkin_span(
+                service_name=self.service_name,
+                span_name=self.span_name,
+                zipkin_attrs=self.zipkin_attrs,
+                transport_handler=self.transport_handler,
+                annotations=self.annotations,
+                binary_annotations=self.binary_annotations,
+                port=self.port,
+                sample_rate=self.sample_rate,
+                include=self.include,
+            ):
                 return f(*args, **kwargs)
         return decorated
 
--- a/tests/zipkin_test.py
+++ b/tests/zipkin_test.py
@@ -458,6 +458,25 @@ def test_zipkin_span_decorator(
     pop_zipkin_attrs_mock.assert_called_once_with()
 
 
+@mock.patch('py_zipkin.zipkin.create_endpoint', wraps=zipkin.create_endpoint)
+def test_zipkin_span_decorator_many(create_endpoint_mock):
+    @zipkin.zipkin_span(service_name='decorator')
+    def test_func(a, b):
+        return a + b
+
+    assert test_func(1, 2) == 3
+    assert create_endpoint_mock.call_count == 0
+    with zipkin.zipkin_span(
+            service_name='context_manager',
+            transport_handler=mock.Mock(),
+            sample_rate=100.0,
+    ):
+        assert test_func(1, 2) == 3
+    assert create_endpoint_mock.call_count == 1
+    assert test_func(1, 2) == 3
+    assert create_endpoint_mock.call_count == 1
+
+
 def test_update_binary_annotations_for_root_span():
     zipkin_attrs = ZipkinAttrs(
         trace_id='0',


py_zipkin/zipkin.py
```diff
@@ -145,6 +145,7 @@ def __init__(
         self.port = port
         self.logging_context = None
         self.sample_rate = sample_rate
+        self.include = include
```
mjbryant: What does this actually do?

py_zipkin/zipkin.py
```diff
@@ -169,7 +170,17 @@ def __init__(
     def __call__(self, f):
         @functools.wraps(f)
         def decorated(*args, **kwargs):
-            with self:
+            with zipkin_span(
+                service_name=self.service_name,
+                span_name=self.span_name,
+                zipkin_attrs=self.zipkin_attrs,
```
mjbryant: Should we be disallowing certain passed-in arguments in this form? In my mind, the decorator form doesn't make any sense for zipkin_attrs or annotations, as these should change on a per-call basis. You're probably never going to want to log span after span with, say, the same trace_id, or the same timestamped annotations.

We could check for `self.zipkin_attrs` and raise an exception if a `zipkin_span` with these set is used in a decorator fashion.

Just a thought.

tests/zipkin_test.py
```diff
@@ -458,6 +458,24 @@ def test_func(a, b):
     pop_zipkin_attrs_mock.assert_called_once_with()
 
 
+@mock.patch('py_zipkin.zipkin.create_endpoint', wraps=zipkin.create_endpoint)
+def test_zipkin_span_decorator_many(create_endpoint_mock):
+    @zipkin.zipkin_span(service_name='decorator')
+    def test_func(a, b):
+        return a + b
+
+    assert create_endpoint_mock.call_count == 0
```
mjbryant: Should you call `test_func` before this assertion?

tests/zipkin_test.py
```diff
@@ -458,6 +458,24 @@ def test_func(a, b):
     pop_zipkin_attrs_mock.assert_called_once_with()
 
 
+@mock.patch('py_zipkin.zipkin.create_endpoint', wraps=zipkin.create_endpoint)
+def test_zipkin_span_decorator_many(create_endpoint_mock):
```
mjbryant: So it seems like you're using `create_endpoint` as a proxy for whether this span tries to log stuff, right? It feels like there should be a better way to do that...

mjbryant:
Re the workaround in #19 - I think we should endeavor to be as defensive as possible, still. I don't mind keeping it.

py_zipkin/zipkin.py
ericchu94: Simply saves the argument so I can pass it in on line 183

py_zipkin/zipkin.py
ericchu94: I do think that the decorator accepts more arguments than it should, but that'll be a backwards incompatible change. We can open up another issue and address it there

tests/zipkin_test.py
ericchu94: oops

tests/zipkin_test.py
ericchu94: The first divergence happens [here](https://github.com/Yelp/py_zipkin/blob/6676f0fdebcab2a3f980ac3a564d1b29e0029752/py_zipkin/zipkin.py#L207), which means I can use `get_zipkin_attrs` to do the same test.

Maybe the more correct way is to do some mock trickery that allows us to inspect the state of the zipkin_span instance, but I can't think of a clean way to do that right now

py_zipkin/zipkin.py
mjbryant: Ohh, of course. I thought this was a new arg to `__init__`. Duh.

py_zipkin/zipkin.py
mjbryant: Yeah, because it'll essentially be broken (or probably result in odd behavior for anyone using those features, at least) for those args. You're right that we should do this in a followup ticket.

tests/zipkin_test.py
mjbryant: Fair nuff

bplotnick:
lg2m. Good work on this!

PR #24, opened 2016-11-17
Add 'py_zipkin.logging_start' default annotation

mjbryant:
@kaisen, @bplotnick what do you think?

### DIFF
--- a/py_zipkin/logging_helper.py
+++ b/py_zipkin/logging_helper.py
@@ -28,7 +28,7 @@ LOGGING_START_KEY = 'py_zipkin.logging_start'
 
 
 class ZipkinLoggingContext(object):
-    """The main logging context manager which controls logging handler and
+    """The main logging context which controls logging handler and
     stores the zipkin attributes on its creation.
 
     :type zipkin_attrs: :class:`py_zipkin.ZipkinAttrs`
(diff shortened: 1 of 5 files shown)


bplotnick:
What's the use case here? I think this sort of measurement is more useful for us as developers of the framework, but I don't see the use to an application developer.

Maybe this would be useful as an option if someone wants to see how much overhead the instrumentation is adding, but I don't know if this is appropriate as a default.

mjbryant:
The use case is to see how much total time is spent in the user's transport_handler. In our case, this is to help us figure out how much we'd gain by switching to a faster transport_handler (such as offloading the actual log generation to a mule). This really feels like something that should be measured. I was looking at some traces earlier and they had these big unaccounted for sections, and I found out that it was due to this. Any other suggestions on ways to do this?

kaisen:
Hmm...do you think the trace would be too noisy if we wrapped client span logging with it's own zipkin_span? Then we could see time spent in transport_handler as a span.

mjbryant:
@kaisen i considered that... i think that's too noisy.

mjbryant:
After some offline discussion, we're going to refactor this to exit the logging context *after* filtering annotations, while adding the 'server' filter to the root span automatically. This will allow us to get rid of adding annotations during logging context manager exit.

For bonus points we can also change the `ZipkinLoggingContext` to not be a contextmanager, since we're not using it as such anyway.

bplotnick:
🚢

kaisen:
:shipit:

bplotnick:
Just for posterity, after _more_ discussion, it was decided to just make the logging annotation optional (which is reflected in the last commit)

Issue #25, opened 2016-11-29
Report span.timestamp and duration


adriancole:
While possible to derive a span's timestamp and duration at query time, it is best to report them to zipkin. This reduces the guesswork on the server side, and particularly makes the cassandra storage implementation more effective. This also makes the duration query operable.

### Do we need to do anything?
Here's the normal doc on that zipkin.io/pages/instrumenting.html This elaborates timestamp and duration a bit, particularly that they can be supplied independently of annotations (sometimes more accurately that way, as a stopwatch isn't at risk of NTP updates)

Then, we'd need to ensure that the special-case of [continuing a client-originated span on the server](https://github.com/openzipkin/openzipkin.github.io/issues/49) is addressed (where B3 headers are read). This might happen here or in pyramid, depending on if we introduce B3 extraction here or not.

Here are tests that should be in place to ensure that span.timestamp and duration are set properly.

* client creates a span and propagates it to a server
  * client propagates b3 headers for trace id, span id, and sampled=1
  * server neither reports span.timestamp nor duration
  * client reports span.timestamp and duration

* server creates a new trace
  * no b3 headers are detected
  * server reports span.timestamp and duration

* server creates a new trace with externally supplied ids
  * caller propagates b3 headers for trace id, span id, but *no sampled header*
  * server reports span.timestamp and duration

Does this make sense? Can anyone here help implement this?

See https://github.com/openzipkin/brave/issues/277#issuecomment-262429392

PR #30, opened 2017-02-01
Properly report timestamp and duration

mjbryant:
Reporting timestamp and duration fields as per http://zipkin.io/pages/instrumenting.html#timestamps-and-duration.

I tried to update stale docstrings and remove unnecessary comments as I went. I also tried to change some (what I thought were, at least) confusing references to a "root" span, as all spans that were called this in the comments weren't truly root spans, in that they weren't the outermost span in a trace. They were simple the outermost span in _the local python process_.

This *may* cause backwards compatibility issues, because the serialized spans are encoded with a different version of the thrift spec. The public python API (i.e. `zipkin_span`) didn't change.

### DIFF
--- a/py_zipkin/logging_helper.py
+++ b/py_zipkin/logging_helper.py
@@ -28,15 +28,13 @@ LOGGING_START_KEY = 'py_zipkin.logging_start'
 
 
 class ZipkinLoggingContext(object):
-    """The main logging context which controls logging handler and
-    stores the zipkin attributes on its creation.
+    """A logging context specific to a Zipkin trace. If the trace is sampled,
+    the logging context sends serialized Zipkin spans to a transport_handler.
+    This is meant to only be used server-side, so the logging context sends
+    a single "server" span, as well as all local child spans collected within
+    the server context.
 
-    :type zipkin_attrs: :class:`py_zipkin.ZipkinAttrs`
-    :type thrift_endpoint: :class:`py_zipkin.thrift.zipkinCore.ttypes.Endpoint`
-    :param log_handler: log handler to be attached to the module logger.
-    :type log_handler: :class:`py_zipkin.logging.ZipkinLoggerHandler`
-    :param span_name:
-    :param transport_handler:
+    This class should only be used by the main `zipkin_span` entrypoint.
     """
 
     def __init__(
@@ -46,6 +44,7 @@ class ZipkinLoggingContext(object):
         log_handler,
         span_name,
         transport_handler,
+        report_root_timestamp,
         binary_annotations=None,
         add_logging_annotation=False,
     ):
@@ -55,6 +54,7 @@ class ZipkinLoggingContext(object):
         self.span_name = span_name
         self.transport_handler = transport_handler
         self.response_status_code = 0
+        self.report_root_timestamp = report_root_timestamp
         self.binary_annotations_dict = binary_annotations or {}
         self.add_logging_annotation = add_logging_annotation
 
@@ -115,6 +115,10 @@ class ZipkinLoggingContext(object):
                 binary_annotations = span['binary_annotations']
                 binary_annotations.update(
                     binary_annotations_by_span_id[span_id])
+
+                timestamp, duration = get_local_span_timestamp_and_duration(
+                    annotations
+                )
                 # Create serializable thrift objects of annotations
                 thrift_annotations = annotation_list_builder(
                     annotations, endpoint
@@ -131,21 +135,25 @@ class ZipkinLoggingContext(object):
                     annotations=thrift_annotations,
                     binary_annotations=thrift_binary_annotations,
                     transport_handler=self.transport_handler,
+                    timestamp_s=timestamp,
+                    duration_s=duration,
                 )
 
-            # Collect extra annotations for server span, then log it.
+            # Collect extra annotations for server span
             extra_annotations = annotations_by_span_id[
                 self.zipkin_attrs.span_id]
             extra_binary_annotations = binary_annotations_by_span_id[
                 self.zipkin_attrs.span_id
             ]
+            server_end_time = time.time()
             annotations = dict(
                 sr=self.start_timestamp,
-                ss=time.time(),
+                ss=server_end_time,
                 **extra_annotations
             )
             if self.add_logging_annotation:
                 annotations[LOGGING_START_KEY] = logging_start
+
             thrift_annotations = annotation_list_builder(
                 annotations,
                 self.thrift_endpoint,
@@ -159,6 +167,12 @@ class ZipkinLoggingContext(object):
                 self.thrift_endpoint,
             )
 
+            if self.report_root_timestamp:
+                timestamp = self.start_timestamp
+                duration = server_end_time - self.start_timestamp
+            else:
+                timestamp = duration = None
+
             log_span(
                 span_id=self.zipkin_attrs.span_id,
                 parent_span_id=self.zipkin_attrs.parent_span_id,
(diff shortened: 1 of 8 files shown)


mjbryant:
@kaisen could you take a look? Also cc @adriancole

py_zipkin/zipkin.py
```diff
@@ -244,7 +245,6 @@ def start(self):
 
         self.start_timestamp = time.time()
 
-        # Set up logging if this is the root span
         if self.perform_logging:
```
mjbryant: Not _quite_. The case where we want to perform logging but don't want to report timestamp/duration is when zipkin_attrs are passed in (implying that this span is a continuation of another span, which implies that the client part of this span will be logged elsewhere and should contain timestamp/duration) and this span isn't independently sampled.

mjbryant:
Addresses #25

py_zipkin/logging_helper.py
```diff
@@ -46,6 +44,7 @@ def __init__(
         log_handler,
         span_name,
         transport_handler,
+        report_root_timestamp,
```
adriancole: ps In Brave (java tracer), I call this flag traceContext.shared

```java
class TraceContext {
--snip--
  public abstract long spanId();

  /**
   * True if we are contributing to a span started by another tracer (ex on a different host).
   * Defaults to false.
   *
   * <p>When an RPC trace is client-originated, it will be sampled and the same span ID is used for
   * the server side. However, the server shouldn't set span.timestamp or duration since it didn't
   * start the span.
   */
  public abstract boolean shared();
```
https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/TraceContext.java#L80

tests/logging_helper_test.py
```diff
@@ -261,15 +282,42 @@ def test_log_span_defensive_about_transport_handler(
...
         annotations='ann',
         binary_annotations='binary_ann',
-        transport_handler=transport_handler,
+        timestamp_s=None,
+        duration_s=None,
+        transport_handler=None,
     )
     assert thrift_obj.call_count == 0
     assert create_sp.call_count == 0
+
```
adriancole: for reference, here's a similar test from the brave java project

```java
  // This prevents the server timestamp from overwriting the client one
  @Test public void doesntReportServerTimestampOnSharedSpans() {
    MutableSpan span = new MutableSpan(context.toBuilder().shared(true).build(), localEndpoint);

    span.start(1L);
    span.kind(SERVER);
    span.finish(2L);

    assertThat(span.toSpan()).extracting(s -> s.timestamp, s -> s.duration)
        .allSatisfy(u -> assertThat(u).isNull());
  }
```

py_zipkin/logging_helper.py
```diff
@@ -159,17 +167,33 @@ def log_spans(self):
                 self.thrift_endpoint,
             )
 
+            if self.report_root_timestamp:
+                timestamp = self.start_timestamp
+                duration = server_end_time - self.start_timestamp
+            else:
```
adriancole: yep in the brave code, looks similar, too, except it uses two signals: shared (ex I didn't start a new trace from an incoming request) and "sr"

```java
    // don't report server-side timestamp on shared or incomplete spans
    if (shared && (flags & FLAG_SR) != 0) {
      span.timestamp(null).duration(null);
    }
```

py_zipkin/thrift/zipkinCore.thrift
```diff
@@ -109,6 +109,68 @@ const string SERVER_RECV_FRAGMENT = "srf"
 
 #***** BinaryAnnotation.key ******
 /**
+ * The domain portion of the URL or host header. Ex. "mybucket.s3.amazonaws.com"
```
adriancole: I presume these changes were taken from https://github.com/openzipkin/zipkin-api/blob/master/thrift/zipkinCore.thrift

py_zipkin/zipkin.py
```diff
@@ -206,16 +206,17 @@ def start(self):
         # If zipkin_attrs are passed in or this span is doing its own sampling,
         # it will need to actually log spans at __exit__.
         self.perform_logging = self.zipkin_attrs or self.sample_rate is not None
+        report_root_timestamp = False
 
         if self.sample_rate is not None:
-            # This clause allows for sampling this service independently
-            # of the passed-in zipkin_attrs.
             if self.zipkin_attrs and not self.zipkin_attrs.is_sampled:
```
adriancole: don't let this confuse you, but in the java impl, we separate out extracting data from the headers, from joining a span or creating a new trace.

```java
contextOrFlags = extractor.extract(request); // the B3 header parser doesn't know if the context will be shared
span = contextOrFlags.context() != null
  ? tracer.joinSpan(contextOrFlags.context()) // this sets shared = true
  : tracer.newTrace(contextOrFlags.samplingFlags()); // this leaves shared = false
```

py_zipkin/zipkin.py
```diff
@@ -290,26 +291,28 @@ def stop(self, _exc_type=None, _exc_value=None, _exc_traceback=None):
...
+        # If we've gotten here, that means that this span is a child span of
+        # this context's root span (i.e. it's a zipkin_span inside another
+        # zipkin_span).
         end_timestamp = time.time()
 
-        # Put the old parent_span_id back on the handler
         self.log_handler.parent_span_id = self.old_parent_span_id
 
-        # To get a full span we just set cs=sr and ss=cr.
+        # We are simulating a full two-part span locally, so set cs=sr and ss=cr
```
adriancole: curious why simulate?

py_zipkin/logging_helper.py
```diff
@@ -250,9 +279,9 @@ def emit(self, record):
                 " to be provided for {0} span".format(span_name)
             )
         service_name = record.msg.get('service_name', None)
-        # Presence of service_name means new client span.
+        # Presence of service_name means this is to be a new local span.
         if service_name is not None:
-            self.store_client_span(
+            self.store_local_span(
```
adriancole: ps a local span wouldn't have sr, ss, cr, cs. There's only one special case, which is that if there are no annotations or binary annotations, we create a fake one (so that indexing works by service name) called "lc"(local component)

ex. in java, we track to see if anything was created prior, and if nothing, we stamp like this..

Ex when there's no annotation, we'd have never added the local endpoint to the span, so do it as a last-ditch attempt
```
    if ((flags & FLAG_LOCAL_ENDPOINT) == 0) { // create a small dummy annotation
      span.addBinaryAnnotation(BinaryAnnotation.create(LOCAL_COMPONENT, "", localEndpoint));
    }
```

py_zipkin/logging_helper.py
mjbryant: Cool. When I was writing this I realized that our current implementation is a bit limited - it doesn't allow generating and logging _only_ a client span. I.e. the outermost `zipkin_span` contextmanager is always going to report ('sr', 'ss').

This is _probably_ okay, since you'll only really ever log client spans in the context of another span. But in this context it means that we don't need to check for 'sr', because the outermost `zipkin_span` (i.e. the one with the logging_context) is always going to log it.

Writing this patch made me want to redesign this library a bit :-/

py_zipkin/thrift/zipkinCore.thrift
mjbryant: Yep

py_zipkin/zipkin.py
mjbryant: Yeah, py-zipkin actually leaves the header extraction to whatever's calling it, e.g. pyramid-zipkin. It just lets you pass in arbitrary headers, which it calls zipkin_attrs, then automatically joins the upstream client span (related to comment above about the outermost `zipkin_span`), always logging ('sr', 'ss').

The brave model is more expressive and complete, but this gets the job done for most sane use cases.

py_zipkin/zipkin.py
mjbryant: In previous iterations we were just logging ('cs', 'cr') for these spans (which is why we call them "client" spans in a lot of the code). I can't recall the exact issue, but we found this didn't play nicely with the UI, so we started logging ('sr', 'ss') as well. So to me logging both client and server annotations felt a bit like cheating - the operation the span was representing is entirely local but the span contained both server and client annotations.

py_zipkin/logging_helper.py
mjbryant: Yeah, see above comment. We just set both client and server annotations due to previous display issues in the UI. It's been working for us so far.

adriancole:
>
> Yeah, see above comment. We just set both client and server annotations
> due to previous display issues in the UI. It's been working for us so far.
>
ok well anyway changing this would be a different topic I suppose. thanks
for the insight.

PR #32, opened 2017-03-03
update_binary_annotations() now handles both root and non-root spans

kaisen:
Previously, zipkin_span had a method called update_binary_annotations_for_root_span which only worked if it was called on a zipkin_span that represented the root span of a trace. In all other cases, it would fail because there would be no logging context associated with the zipkin_span.

It turns out that all one has to do to add binary_annotations to a non-root span is to just update the self.binary_annotations dictionary directly, but it's not immediately obvious. By changing the update function to be called update_binary_annotations() and making it handle both root and non-root spans, you can now just call it on a span context and expect it to work correctly

This will require a pyramid_zipkin change as well.

To clarify the before/after effects of this PR...

BEFORE:
```python
with zipkin_span(
    service_name='some_service',
    span_name='a_root_span',
    sample_rate=100.0,
) as zipkin_context:
    # can use update_binary_annotations_for_root_span() to add annotations to the root span
    zipkin_context.update_binary_annotations_for_root_span({'some_key': 'some_val'})

    with zipkin_span(
        service_name='some_service',
        span_name='nested_span',
    ) as nested_zipkin_span:
        # have to update binary annotations directly
        nested_zipkin_span.binary_annotations.update({'some_key': 'some_val'})
```

Having 2 different methods of updating binary annotations based on whether the span is a root span or not is confusing. This PR just consolidates both methods into a single update_binary_annotations method.

### DIFF
--- a/tests/zipkin_test.py
+++ b/tests/zipkin_test.py
@@ -533,7 +533,7 @@ def test_update_binary_annotations():
             assert 'two' not in context.logging_context.binary_annotations_dict
 
 
-def test_update_binary_annotations_errors():
+def test_update_binary_annotations_should_not_error_if_not_tracing():
     zipkin_attrs = ZipkinAttrs(
         trace_id='0',
         span_id='1',
--- a/tests/zipkin_test.py
+++ b/tests/zipkin_test.py
@@ -500,7 +500,7 @@ def test_zipkin_span_add_logging_annotation(mock_context):
     assert kwargs['add_logging_annotation']
 
 
-def test_update_binary_annotations_for_root_span():
+def test_update_binary_annotations():
     zipkin_attrs = ZipkinAttrs(
         trace_id='0',
         span_id='1',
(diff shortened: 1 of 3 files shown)


tests/zipkin_test.py
```diff
@@ -518,11 +518,22 @@ def test_update_binary_annotations_for_root_span():
...
+        )
+        with nested_context:
+            assert 'one' not in context.logging_context.binary_annotations_dict
+            nested_context.update_binary_annotations({'two': 'two'})
+            assert 'two' in nested_context.binary_annotations
+            assert 'two' not in context.logging_context.binary_annotations_dict
+
 
-def test_update_binary_annotations_for_root_span_errors():
+def test_update_binary_annotations_errors():
```
kaisen: test name updated

kaisen:
Added a more explicit example to the original PR description.

bplotnick:
⛵️ !

PR #34, opened 2017-04-27
Check for no zipkin_attrs when updating binary annotations

kaisen:
Calling update_binary_annotations on a context that doesn't have zipkin_attrs (which occurs when the span is a child span in a trace that is not being sampled) results in an attribute error. This fixes the bug by also checking for the existence of zipkin_attrs inside update_binary_annotations.


bplotnick:
👍

Issue #28, opened 2017-01-11
Support 128-bit trace IDs


mjbryant:
https://github.com/openzipkin/zipkin/releases/tag/1.14.4 added support for 128-bit trace IDs via a new `trace_id_high` field in the thrift schema. We'll have to update to support this use case.

imcom:
Hi @mjbryant how is this going so far ? is there a schedule to add support of 128bit trace IDs ?

mjbryant:
This likely isn't being worked on. @kaisen might have more context.

bplotnick:
This was fixed in #36

adriancole:
thanks!

PR #36, opened 2017-05-17
Add 128-bit trace id support

kaisen:
#28 

A `use_128bit_trace_id` flag on zipkin_span will control whether or not 128-bit (32 char hext string) trace ids are generated. Default is False.

### DIFF
--- a/py_zipkin/thrift/__init__.py
+++ b/py_zipkin/thrift/__init__.py
@@ -143,7 +143,7 @@ def create_span(
     trace_id_high = None
     if trace_id_length > 16:
         assert trace_id_length == 32
-        trace_id, trace_id_high = trace_id[:16], trace_id[16:]
+        trace_id, trace_id_high = trace_id[16:], trace_id[:16]
 
     span_dict = {
         'trace_id': unsigned_hex_to_signed_int(trace_id),
--- a/py_zipkin/thrift/__init__.py
+++ b/py_zipkin/thrift/__init__.py
@@ -153,8 +153,8 @@ def create_span(
         'binary_annotations': binary_annotations,
         'timestamp': int(timestamp_s * 1000000) if timestamp_s else None,
         'duration': int(duration_s * 1000000) if duration_s else None,
-        'trace_id_high': unsigned_hex_to_signed_int(trace_id_high) \
-            if trace_id_high else None,
+        'trace_id_high': unsigned_hex_to_signed_int(trace_id_high)
+        if trace_id_high else None,
     }
     if parent_span_id:
         span_dict['parent_id'] = unsigned_hex_to_signed_int(parent_span_id)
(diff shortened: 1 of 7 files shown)


py_zipkin/thrift/__init__.py
```diff
@@ -138,6 +138,13 @@ def create_span(
     of the span. Timestamps passed in are in seconds, they're converted to
     microseconds before thrift encoding.
     """
+    # Check if trace_id is 128-bit. If so, record trace_id_high separately.
+    trace_id_length = len(trace_id)
+    trace_id_high = None
+    if trace_id_length > 16:
+        assert trace_id_length == 32
+        trace_id, trace_id_high = trace_id[:16], trace_id[16:]
```
bplotnick: I think you have this backwards. The high order bytes are on the left side of the string. See the example here: https://github.com/openzipkin/zipkin/blob/master/zipkin-server/README.md#128-bit-trace-ids

tests/thrift/thrift_test.py
```diff
@@ -24,6 +25,48 @@ def test_create_span():
...
+        parent_span_id='0000000000000002',
+        trace_id='000000000000000f000000000000000f',
+        span_name='foo',
+        annotations='ann',
+        binary_annotations='binary_ann',
+        timestamp_s=1485920381.2,
+        duration_s=2.0,
+    )
+    assert span.trace_id == 15
+    assert span.trace_id_high == 15
```
bplotnick: Can you have the the high order portion and low order portion be different so we can be sure that the extraction is correct?

py_zipkin/thrift/__init__.py
```diff
@@ -146,6 +153,8 @@ def create_span(
         'binary_annotations': binary_annotations,
         'timestamp': int(timestamp_s * 1000000) if timestamp_s else None,
         'duration': int(duration_s * 1000000) if duration_s else None,
+        'trace_id_high': unsigned_hex_to_signed_int(trace_id_high)
+        if trace_id_high else None,
```
bplotnick: i'm not sure if the the linter formatted it this way, but i would've indented this line more to make it clearer that it is part of the above line

py_zipkin/zipkin.py
```diff
@@ -406,7 +413,12 @@ def __init__(self, *args, **kwargs):
         super(zipkin_server_span, self).__init__(*args, **kwargs)
 
 
-def create_attrs_for_span(sample_rate=100.0, trace_id=None, span_id=None):
+def create_attrs_for_span(
+    sample_rate=100.0,
+    trace_id=None,
+    span_id=None,
+    use_128bit_trace_id=False
```
bplotnick: nit: trailing comma

py_zipkin/thrift/__init__.py
bplotnick: oh I see now you actually did in the initial version.  👎  flake8. It's still kind of hard to read, so maybe just create this in a temporary variable before the span_dict line?

adriancole:
Zipkin has a compatibility mode such that if there are mixed spans in the
same trace, queries ignore the high bits.

bplotnick:
🚢

PR #38, opened 2017-05-25
Modified zipkin_span to include host variable

DevelopingCoder:


py_zipkin/zipkin.py
```diff
@@ -109,6 +109,7 @@ def __init__(
         add_logging_annotation=False,
         report_root_timestamp=False,
         use_128bit_trace_id=False,
+        host=None
```
kaisen: add a docstring explaining that this will expect the ipv4 value of the host and that the ipv4 value won't be automatically determined

py_zipkin/zipkin.py
```diff
@@ -266,8 +268,7 @@ def start(self):
             # Don't set up any logging if we're not sampling
             if not self.zipkin_attrs.is_sampled:
                 return self
-
-            endpoint = create_endpoint(self.port, self.service_name)
+            endpoint = create_endpoint(self.port, self.service_name, self.host)
```
kaisen: this will require a test to make sure that host overriding will work :)

kaisen:
:shipit:

PR #39, opened 2017-05-30
handle socket exception in create_endpoint

### LINKED ISSUE #37: If your host can not be resolved by DNS, it will break tracing
In some cases a machine host cannot be successfully resolved be DNS (for instance, dev machines weren't fully setup in my case). If your host is not resolvable, you will get an exception in  https://github.com/Yelp/py_zipkin/blob/master/py_zipkin/thrift/__init__.py#L64. This problem fully breaks tracing. I think this can be processed in more graceful way. May be it's better to log and swallow the exception and use the hostname as an endpoint. What do you think, guys?

bplotnick:
This fixes #37 by catching `socket.gaierror` (the exception for an unresolvable name) and defaults to `127.0.0.1`.

I didn't add any logging for the error, but i think it might be nice. @kaisen, is there currently a preferred way to log warnings?

### DIFF
--- a/py_zipkin/thrift/__init__.py
+++ b/py_zipkin/thrift/__init__.py
@@ -61,7 +61,10 @@ def create_endpoint(port=0, service_name='unknown', host=None):
     :returns: zipkin Endpoint object
     """
     if host is None:
-        host = socket.gethostbyname(socket.gethostname())
+        try:
+            host = socket.gethostbyname(socket.gethostname())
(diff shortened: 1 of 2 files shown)


kaisen:
:shipit:

bplotnick:
It looks like there's no real way to do the warning logging right now. I filed #40 to do that. This can be logged as a warning once that's done.

Issue #41, opened 2017-05-30
Support all available annotation types for binary annotations


kaisen:
According to the zipkin thrift spec, binary annotations can be of types bool, bytes, int, double, and string. py_zipkin currently only supports string binary annotations.

adriancole:
I'd prefer if you don't implement this. It isn't likely the other types
will carry forward, and zipkin really only supports string.
https://github.com/openzipkin/zipkin/issues/1499

kaisen:
Thanks @adriancole, good to know!

PR #43, opened 2017-05-31
Add support for adding a 'sa' binary annotation to spans

### LINKED ISSUE #33: Log 'sa' annotation for client spans
This is from zipkin docs (http://zipkin.io/pages/data_model.html)
> spans that include cs can log an sa annotation of where they are going
> This helps when the destination protocol isn’t Zipkin instrumented, such as MySQL.

It seems that you can't log third-party server address using in such way. Please advise.

kaisen:
This resolves #33.

Note that I chose to save the 'sa' binary annotation as its own unique property instead of putting it into the zipkin_span.binary_annotations property (because that property only supports dicts of str -> str right now). #41 will require a change in zipkin_span.binary_annotations, so I'm holding off on that work for now.

### DIFF
--- a/py_zipkin/logging_helper.py
+++ b/py_zipkin/logging_helper.py
@@ -56,6 +56,7 @@ class ZipkinLoggingContext(object):
         self.response_status_code = 0
         self.report_root_timestamp = report_root_timestamp
         self.binary_annotations_dict = binary_annotations or {}
+        self.sa_binary_annotation = None
         self.add_logging_annotation = add_logging_annotation
 
     def start(self):
@@ -126,6 +127,8 @@ class ZipkinLoggingContext(object):
                 thrift_binary_annotations = binary_annotation_list_builder(
                     binary_annotations, endpoint
                 )
+                if 'sa_binary_annotation' in span and span['sa_binary_annotation']:
+                    thrift_binary_annotations.append(span['sa_binary_annotation'])
 
                 log_span(
                     span_id=span_id,
@@ -166,6 +169,8 @@ class ZipkinLoggingContext(object):
                 self.binary_annotations_dict,
                 self.thrift_endpoint,
             )
+            if self.sa_binary_annotation:
+                thrift_binary_annotations.append(self.sa_binary_annotation)
 
             if self.report_root_timestamp:
                 timestamp = self.start_timestamp
@@ -219,6 +224,7 @@ class ZipkinLoggerHandler(logging.StreamHandler, object):
         service_name,
         annotations,
         binary_annotations,
+        sa_binary_annotation=None,
         span_id=None,
     ):
         """Convenience method for storing a local child span (a zipkin_span
@@ -232,6 +238,7 @@ class ZipkinLoggerHandler(logging.StreamHandler, object):
             'span_id': span_id,
             'annotations': annotations,
             'binary_annotations': binary_annotations,
+            'sa_binary_annotation': sa_binary_annotation,
         })
 
     def emit(self, record):
--- a/py_zipkin/logging_helper.py
+++ b/py_zipkin/logging_helper.py
@@ -56,7 +56,7 @@ class ZipkinLoggingContext(object):
         self.response_status_code = 0
         self.report_root_timestamp = report_root_timestamp
         self.binary_annotations_dict = binary_annotations or {}
-        self.sa_binary_annotation = None
+        self.sa_binary_annotations = []
         self.add_logging_annotation = add_logging_annotation
 
     def start(self):
@@ -127,8 +127,8 @@ class ZipkinLoggingContext(object):
                 thrift_binary_annotations = binary_annotation_list_builder(
                     binary_annotations, endpoint
                 )
-                if 'sa_binary_annotation' in span and span['sa_binary_annotation']:
-                    thrift_binary_annotations.append(span['sa_binary_annotation'])
+                if span.get('sa_binary_annotations'):
+                    thrift_binary_annotations += span['sa_binary_annotations']
 
                 log_span(
                     span_id=span_id,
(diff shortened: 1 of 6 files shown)


py_zipkin/logging_helper.py
```diff
@@ -126,6 +127,8 @@ def log_spans(self):
                 thrift_binary_annotations = binary_annotation_list_builder(
                     binary_annotations, endpoint
                 )
+                if 'sa_binary_annotation' in span and span['sa_binary_annotation']:
```
bplotnick: `if span.get('sa_binary_annotation'):` might be more readable

py_zipkin/zipkin.py
```diff
@@ -375,6 +383,50 @@ def update_binary_annotations(self, extra_annotations):
...
+            return
+
+        sa_endpoint = create_endpoint(
+            port=port,
+            service_name=service_name,
+            host=host,
+        )
+        sa_binary_annotation = create_binary_annotation(
+            key='sa',
+            value=str(True),
```
bplotnick: I don't think the value should be stringified. It is a boolean type

py_zipkin/zipkin.py
```diff
@@ -375,6 +383,50 @@ def update_binary_annotations(self, extra_annotations):
...
+            annotation_type=zipkin_core.AnnotationType.BOOL,
+            host=sa_endpoint,
+        )
+        if not self.logging_context:
+            self.sa_binary_annotation = sa_binary_annotation
+        else:
+            self.logging_context.sa_binary_annotation = sa_binary_annotation
```
bplotnick: According to the spec, there can be multiple `sa` or `ca` annotations, so you'll probably want to make `sa_binary_annotation` into a list and pluralize the variable name.

py_zipkin/zipkin.py
```diff
@@ -375,6 +383,50 @@ def update_binary_annotations(self, extra_annotations):
...
+            # should result in a logged error
+            return
+
+        sa_endpoint = create_endpoint(
+            port=port,
+            service_name=service_name,
+            host=host,
+        )
+        sa_binary_annotation = create_binary_annotation(
+            key='sa',
```
bplotnick: We don't do this elsewhere in the code (we should), but can you use `zipkin_core.SERVER_ADDR` instead? I think it's a bit clearer

py_zipkin/zipkin.py
kaisen: the thrift spec says that this field is supposed to be a byte string. thriftpy will encode this value for us, but it must be a str-like object.

kaisen:
I'll hold off 'ca' stuff for another PR.

py_zipkin/thrift/__init__.py
```diff
@@ -13,6 +13,7 @@
 thrift_filepath = os.path.join(os.path.dirname(__file__), 'zipkinCore.thrift')
 zipkin_core = thriftpy.load(thrift_filepath, module_name="zipkinCore_thrift")
 
+SERVER_ADDR_VAL = '\x01'
```
bplotnick: Will this work in python 3 without the `b` prefix? I still don't fully understand what's going on here. Combine that with my general lack of understanding of strings in python3 and this question is probably dumb.

kaisen: i tested it in py3 and both values result in the same thriftpy object in the end.

bplotnick:
woot! 🚢

dmitry-prokopchenkov:
@kaisen Looks like py_zipkin doesn't allow to use dns addresses as endpoints (and 'sa' endpoints accordingly). For instance, to create 'sa' endpoint we have to pass ipv4 address in 'host' param https://github.com/Yelp/py_zipkin/blob/master/py_zipkin/zipkin.py#L391 or it will fail at https://github.com/Yelp/py_zipkin/blob/master/py_zipkin/thrift/__init__.py#L70. Could you please explain this restriction.

adriancole:
This is likely because of direct mapping to wire format. Also, it helps
from adding overhead to a request by doing DNS lookups.

Note: in java (brave) we recently added helpers for this, particularly as
it also helps support ipv6. Notice that this is an option, as you can still
pass IP binary literals.

public boolean parseIp(@Nullable String ipString) {
  if (ipString == null) return false;
  byte[] addressBytes = InetAddresses.ipStringToBytes(ipString);
  if (addressBytes == null) return false;
  if (addressBytes.length == 4) {
    ipv4(ByteBuffer.wrap(addressBytes).getInt());
  } else if (addressBytes.length == 16) {
    ipv6(addressBytes);
  } else {
    return false;
  }
  return true;
}

dmitry-prokopchenkov:
@adriancole I suppose that 'sa' annotation is just a common  place where we can store server address, if one service uses DNS address to access other service, then we should be able to place this DNS in 'sa', don't they?

@kaisen also I would like to get your comments

adriancole:
>
> @adriancole <https://github.com/adriancole> I suppose that 'sa'
> annotation is just a common place where we can store server address, if one
> service uses DNS address to access other service, then we should be able to
> place this DNS in 'sa', don't they?
>
"sa" is special as the value is stored in the Endpoint "binary annotation"
(unlike normal tags where the value is a string). This is defined in thrift
and only includes a resolved ipv4 or ipv6 ip or port (if present). Often, a
client interceptor or callback can show the actual IP address used for the
connection. If not, sometimes people resolve DNS, but that could be tricky.
Often it is better to add the tag/binary annotation "http.host" separately
if you don't know the actual IP address, but have some host info.

So, "sa" should be a binary annotation with key = "sa", type=bool, value=[1]
endpoint.serviceName=name_of_remote_service
endpoint.(ipv4|ipv6)=exact_ip_used or absent if not known
endpoint.port=exact_port_used or absent if not known

https://github.com/openzipkin/zipkin-api/blob/master/thrift/zipkinCore.thrift

In the future, this will be simplified to span.remoteService with exactly
the same data

Issue #42, opened 2017-05-31
server send timings overstated


Jefftree:
In a few zipkin traces, there were instances where server send was being emitted after client receive. It seems that "server send" represents the time when zipkin logging is finished (for a root span). However, a function being traced would already be completed when zipkin logging starts.

I believe the "server send" should be timestamped at the same time as `logging_start` since `logging_start` occurs when a function ends, and waiting until logging is completed would skew the timings (and make the span appear longer than the amount of time it actually took). We could potentially add a `logging_end` field to where the current "server send" is being logged.

trtg:
:+1:  locally this issue didn't seem to be that huge, but now that i'm testing py_zipkin in a production/high-volume and distributed environment, the current duration of the root span is wildly overestimated (marked as >= 3 seconds) when all the nested spans only take up < 200ms. If I change ss to be logging_start, and duration to be logging_start - self.start_timestamp, the root span is only as wide as all the nested spans.
Also, this over estimate of duration happens every single time in production, it's not an occasional issue.

kaisen:
@trtg fantastic, thanks for helping us confirm that this is an issue that needs to be addressed!

bplotnick:
This is has been done since #47

PR #47, opened 2017-06-01
Change server send timing to logging_start

Jefftree:
Issue Link: https://github.com/Yelp/py_zipkin/issues/42

Fixes server send timing to more accurately reflect when server send actually occurs.

Removed the `logging_start` annotation (since it'll be the same as Server Send) and added a `logging_end` annotation

### DIFF
--- a/py_zipkin/logging_helper.py
+++ b/py_zipkin/logging_helper.py
@@ -151,7 +151,7 @@ class ZipkinLoggingContext(object):
             server_end_time = time.time()
             annotations = dict(
                 sr=self.start_timestamp,
-                ss=server_end_time,
+                ss=logging_start,
                 **extra_annotations
             )
             if self.add_logging_annotation:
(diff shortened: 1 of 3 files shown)


bplotnick:
Yes, I believe the root span duration should be changed as well (I'm curious as to how the zipkin UI will render this).

Also, you will need to remove `logging_end` and probably add `logging_end` in place of where `ss` used to be.

Jefftree:
@bplotnick We current don't have a field for `logging_end`, do you mean adding an extra annotation `logging_end`?

bplotnick:
Whoops. I meant remove `logging_start` and add `logging_end`. And yeah I mean add it as an annotation (basically what we did for `logging_start`)

bplotnick:
Can you also update the docstring for `zipkin.py:zipkin_span.__init__`?

Other than that, it looks good to me.

bplotnick:
:shipit:

kaisen:
👌 🚢

PR #48, opened 2017-06-21
Add error binary annotation if the context manager exits with an exception

kaisen:


kaisen:
Are we worried about the binary annotation value containing the stacktrace being too large? It's going to be as large as a standard python stacktrace. The most I've seen in production systems is a 30ish-line stack trace which isn't unreasonable.

I feel that this is more of a problem for the zipkin UI to handle. IMO having the full stacktrace is useful in debugging in the event of an error annotation being logged. The value is also stored in json format making it more flexible to deal with.

adriancole:
The problem I was raising was more about the data being created is tested
with mocks so hard to reason with. There are large annotations in zipkin
sometimes a kilobyte or more, but the way this is tested, i have no idea if
it is large or not.. hence the question.

Large annotations end up a system concern not just a UI one due to making
things like kafka message size a problem. It doesnt seem like this is the
case here, just yeah at least an example of what text this code produces
would eliminate this concern (even ok if there is an assertion like less
than 1KiB)

Definitely a span needs to be closed so adding duration or cr on error is
important.

adriancole:
Ps on system concern.. ex in finagle all spans are basically the same size.
This makes scaling and storage math easy as a rate of spans/time unit.

Thats why in brave we default to have error as one line message so easy to
reason with. We have a hook to allow folks to opt into making large tags,
but default is less.

Ack over-explaining, just this sort of stuff might not be obvious. No
problem here just mentioning some things to consider whatever direction
chosen.

On 23 Jun 2017 8:30 am, wrote:

The problem I was raising was more about the data being created is tested
with mocks so hard to reason with. There are large annotations in zipkin
sometimes a kilobyte or more, but the way this is tested, i have no idea if
it is large or not.. hence the question.

Large annotations end up a system concern not just a UI one due to making
things like kafka message size a problem. It doesnt seem like this is the
case here, just yeah at least an example of what text this code produces
would eliminate this concern (even ok if there is an assertion like less
than 1KiB)

Definitely a span needs to be closed so adding duration or cr on error is
important.

bplotnick:
🚢

Issue #35, opened 2017-05-14
How can I create client root span?


dmitry-prokopchenkov:
'sr' and 'ss' annotations are always added to root span in [logging_helper.py:150](https://github.com/Yelp/py_zipkin/blob/master/py_zipkin/logging_helper.py#L150). It's not correctly for client root spans.

adriancole:
sounds like a legit limitation in the current code.

dmitry-prokopchenkov:
Guys, would you mind if I create a pull request and add this ability by myself?

bplotnick:
@dmitry-prokopchenkov by all means!! We love pull requests! I'm actually not sure how this would look, so i'm interested to see what you might come up with 😄 . Let us know if you need a hand.

P.S. For some historical context, the reason this limitation exists is that py_zipkin was formed out of our initial project pyramid_zipkin, and in fact is still the main way we use py_zipkin. For us, everything starts with a service request. But I totally see why it would be useful to be able to start with a client span.

dmitry-prokopchenkov:
ok, thanks, I will give it a try when I have some time

bplotnick:
closed via #49

kaisen:
@dmitry-prokopchenkov your change is included in the tagged v0.8.2 release :)

dmitry-prokopchenkov:
@kaisen sorry, it was better to not change the release version, yes?

bplotnick:
@dmitry-prokopchenkov i'm not sure what you're asking. are you talking about why we bumped the patch version rather than the minor version? we generally use semver at Yelp, but releases <1.0.0 have unstable APIs, so the usual semver rules are much looser and it's kind of a judgement call.

kaisen:
@dmitry-prokopchenkov did you mean that it would have been better if you didn't change the release version in your PR? It doesn't matter to me :)

dmitry-prokopchenkov:
@kaisen , @bplotnick ok, thanks for the explanations. I am looking forward for answer on my [question](https://github.com/Yelp/py_zipkin/issues/6#issuecomment-312414780) about batch span sending. This is that I might add quickly enough.

PR #49, opened 2017-06-22
Added processing for client root spans

dmitry-prokopchenkov:
This is pull request for https://github.com/Yelp/py_zipkin/issues/35

### DIFF
--- a/py_zipkin/logging_helper.py
+++ b/py_zipkin/logging_helper.py
@@ -30,10 +30,8 @@ LOGGING_END_KEY = 'py_zipkin.logging_end'
 class ZipkinLoggingContext(object):
     """A logging context specific to a Zipkin trace. If the trace is sampled,
     the logging context sends serialized Zipkin spans to a transport_handler.
-
-    The trace can represent
-        1) server and contains root server span and child client spans (optionally)
-        2) client and contains root client span only
+    The logging context sends root "server" or "client" span, as well as all
+    local child spans collected within this context.
 
     This class should only be used by the main `zipkin_span` entrypoint.
     """
@@ -101,7 +99,6 @@ class ZipkinLoggingContext(object):
                 )
 
             # Collect, annotate, and log client spans from the logging handler
-            # In case of root client span client_spans list is empty
             for span in self.log_handler.client_spans:
                 # The parent_span_id is either the parent ID set in the
                 # logging handler or the current Zipkin context's span ID.
--- a/py_zipkin/logging_helper.py
+++ b/py_zipkin/logging_helper.py
@@ -30,9 +30,10 @@ LOGGING_END_KEY = 'py_zipkin.logging_end'
 class ZipkinLoggingContext(object):
     """A logging context specific to a Zipkin trace. If the trace is sampled,
     the logging context sends serialized Zipkin spans to a transport_handler.
-    This is meant to only be used server-side, so the logging context sends
-    a single "server" span, as well as all local child spans collected within
-    the server context.
+
+    The trace can represent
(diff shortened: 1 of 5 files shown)


py_zipkin/logging_helper.py
```diff
@@ -30,9 +30,10 @@ def emit(self, record):
 class ZipkinLoggingContext(object):
     """A logging context specific to a Zipkin trace. If the trace is sampled,
     the logging context sends serialized Zipkin spans to a transport_handler.
-    This is meant to only be used server-side, so the logging context sends
-    a single "server" span, as well as all local child spans collected within
-    the server context.
+
+    The trace can represent
+        1) server and contains root server span and child client spans (optionally)
+        2) client and contains root client span only
```
bplotnick: I think this new comment is not quite correct. The original comment was saying basically that the root span is a `server` span, but the local child spans can be client and/or server. In fact, our "local" spans use both cs/cr and ss/sr annotations.

You can still have this with a client root span. Imagine a application that will download some file, and also has some local method for validation that we instrumented.  We might have

```
cs:          download_file
cs/sr/ss/sr: validate_parameters_locally
cs:          fetch_file_from_server
cr:          fetch_file_from_server
cr:          download_file
```

You could even imagine child server spans in there if somewhere in the middle we check some queue for new data.

All of this is pretty unfamiliar, so i'm open to any other opinions on this matter.

py_zipkin/logging_helper.py
```diff
@@ -97,6 +101,7 @@ def log_spans(self):
                 )
 
             # Collect, annotate, and log client spans from the logging handler
+            # In case of root client span client_spans list is empty
```
bplotnick: same comment as above, I don't think that this is quite correct

py_zipkin/zipkin.py
```diff
@@ -281,6 +281,7 @@ def start(self):
             if not self.zipkin_attrs.is_sampled:
                 return self
             endpoint = create_endpoint(self.port, self.service_name, self.host)
+            client_context = (self.include == {'client'})
```
dmitry-prokopchenkov: thanks!

py_zipkin/logging_helper.py
dmitry-prokopchenkov: totally agree with you, will change the comment

py_zipkin/logging_helper.py
dmitry-prokopchenkov: I removed my comment. May be it would be better to rename client_spans (self.log_handler.client_spans) to child_spans?

bplotnick: I'd be totally cool with that if you'd like to tack that on to this PR. Otherwise, I can do that as a followup task.

dmitry-prokopchenkov: The second option, I can do that, but I would like to do it separately.

bplotnick:
looks good to me!

@kaisen can you take a quick look and see if anything pops out to you?

dmitry-prokopchenkov:
@bplotnick I am not good familiar with github pull request workflow. Should I close this pull request? Or at first wait for @kaisen's answer?

bplotnick:
@dmitry-prokopchenkov once @kaisen looks at it (i just poked him), assuming he has no objection, i will go ahead and merge your PR

kaisen:
nice! looks good

bplotnick:
@dmitry-prokopchenkov this has been merged. Thanks so much for this!

dmitry-prokopchenkov:
thank you, guys!

PR #51, opened 2017-07-09
Be more defensive about logging handlers being configured

bplotnick:
We already had some [logic](https://github.com/Yelp/py_zipkin/pull/51/files#diff-bc19fbb857d40c54c3d763ce4021abe8R304) to be defensive about the fact that logging handlers may not be configured in multithreaded applications. However, this logic has been incorrect since the NullHandler was [added](https://github.com/Yelp/pyramid_zipkin/commit/6c5adb294372403c5ef5a6d4c835c7b282b3c216).


adriancole:
LG!

PR #52, opened 2017-07-18
Added batch spans sending

dmitry-prokopchenkov:
This is a pr for https://github.com/Yelp/py_zipkin/issues/6

Guys, I have a question. I didn't provide a backward compatibility in this pull request. After my changes in thrift encoding code this example from the docs won't work:
```
import requests

def http_transport(encoded_span):
    # The collector expects a thrift-encoded list of spans. Instead of
    # decoding and re-encoding the already thrift-encoded message, we can just
    # add header bytes that specify that what follows is a list of length 1.
    body = '\x0c\x00\x00\x00\x01' + encoded_span
    requests.post(
        'http://localhost:9411/api/v1/spans',
        data=body,
        headers={'Content-Type': 'application/x-thrift'},
    )
```
I think here we breach encapsulation of thrift encoding logic by adding '\x0c\x00\x00\x00\x01'. In the pull request the count of thrift objects is defined automatically and can be hided from client. So the modified version of this example would be:
```
import requests

def http_transport(encoded_span):
    requests.post(
        'http://localhost:9411/api/v1/spans',
        data=encoded_span,
        headers={'Content-Type': 'application/x-thrift'},
    )
```

I can provide a backward compatibility by adding a 'batch' flag to zipkin_span constructor, but I prefer not to do it without your feedback. Please advise.

### DIFF
--- a/py_zipkin/logging_helper.py
+++ b/py_zipkin/logging_helper.py
@@ -46,7 +46,8 @@ class ZipkinLoggingContext(object):
         report_root_timestamp,
         binary_annotations=None,
         add_logging_annotation=False,
-        client_context=False
+        client_context=False,
+        max_span_portion_size=None,
     ):
         self.zipkin_attrs = zipkin_attrs
         self.thrift_endpoint = thrift_endpoint
@@ -59,6 +60,7 @@ class ZipkinLoggingContext(object):
         self.sa_binary_annotations = []
         self.add_logging_annotation = add_logging_annotation
         self.client_context = client_context
+        self.max_span_portion_size = max_span_portion_size
 
     def start(self):
         """Actions to be taken before request is handled.
@@ -85,8 +87,12 @@ class ZipkinLoggingContext(object):
         a success. It also logs the service (`ss` and `sr`) or the client
         ('cs' and 'cr') annotations.
         """
-        if self.zipkin_attrs.is_sampled:
-            span_sender = ZipkinBatchSender(self.transport_handler)
+        if not self.zipkin_attrs.is_sampled:
+            return
+
+        span_sender = ZipkinBatchSender(self.transport_handler,
+                                        self.max_span_portion_size)
+        with span_sender:
             end_timestamp = time.time()
             # Collect additional annotations from the logging handler
             annotations_by_span_id = defaultdict(dict)
@@ -190,7 +196,6 @@ class ZipkinLoggingContext(object):
                 timestamp_s=timestamp,
                 duration_s=duration,
             )
-            span_sender.flush()
 
 
 def get_local_span_timestamp_and_duration(annotations):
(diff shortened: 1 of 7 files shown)


kaisen:
@bplotnick if I'm reading the code right, it looks like we won't have to worry about spans not being sent because flush() is always called with the logging context manager exits. Unless you're thinking about another corner case?

Still reviewing this PR.

bplotnick:
@kaisen My concern was not with `log_spans`, but rather if someone uses `ZipkinBatchSender` separately. Perhaps we can call flush in the `ZipkinBatchSender` destructor?

dmitry-prokopchenkov:
@bplotnick re: flush(). I don't think zipkin-reporter-java with its timeouts is suitable for our case, because here we encode all spans together in zipkin_span.stop(). But we can use ZipkinBatchSender as a context manager and do flush() at \_\_exit__().

dmitry-prokopchenkov:
Guys, looking forward to your feedback about the latest changes

py_zipkin/logging_helper.py
```diff
@@ -329,6 +338,29 @@ def log_span(
         binary_annotations,
         timestamp_s,
         duration_s,
-    )
-    message = thrift_obj_in_bytes(span)
-    transport_handler(message)
+    ):
+        if not self.transport_handler:
```
kaisen: i believe this check is unnecessary. a logging context is only created if perform_logging is True, which is only True if (self.zipkin_attrs or self.sampling_rate is not None) which requires a check for self.transport_handler.

py_zipkin/zipkin.py
```diff
@@ -126,6 +127,9 @@ def __init__(
         :param transport_handler: Callback function that takes a message parameter
             and handles logging it
         :type transport_handler: function
+        :param max_span_portion_size: Spans in a trace are sent in batches,
```
kaisen: doesn't matter too much but i would prefer 'max_span_batch_size'. @bplotnick thoughts?

tests/integration/zipkin_integration_test.py
```diff
@@ -1,5 +1,5 @@
 import pytest
-from thriftpy.protocol.binary import TBinaryProtocol
+from thriftpy.protocol.binary import TBinaryProtocol, read_list_begin
```
kaisen: 2 separate import lines for this

py_zipkin/zipkin.py
bplotnick: Yeah I like the idea of having "batch" somewhere in the name, considering the batch sender class is called `ZipkinBatchSender`

dmitry-prokopchenkov:
@kaisen, @bplotnick Please verify my latest commit. I took into account all your comments.

dmitry-prokopchenkov:
Guys?

kaisen:
Sorry for the delay @dmitry-prokopchenkov. This looks good to me. @bplotnick ?

bplotnick:
:shipit:

bplotnick:
Merged. I'll release a 0.9.0 in a bit with these changes. Thanks so much for this work @dmitry-prokopchenkov!!

dmitry-prokopchenkov:
@bplotnick thanks! I really need this version in pip to complete my current task) When are you going to release this?

bplotnick:
@kaisen I released 0.9.0 yesterday, but it didn't get uploaded to the public pypi (for some reason, i thought we had this automated). Can you do this so @dmitry-prokopchenkov can use the release?

bplotnick:
@dmitry-prokopchenkov v0.9.0 is now on pypi

dmitry-prokopchenkov:
@bplotnick , @kaisen Thanks!

PR #63, opened 2017-11-28
Force .perform_logging to bool

topiaruss:
Without this change, perform_logging ends up with the the value of any passed zipkin_attrs.


kaisen:
Thanks for the PR! I'm probably going to merge this, but I'm curious: is this fixing a bug? From what I see, `self.perform_logging` is only evaluated for truthiness, so the code should work regardless of the actual type of `self.perform_logging`.

topiaruss:
I think you are probably right. And I'm not aware of a bug.
I was debugging an issue and surprised to see a list in what looked like a boolean field.
Just cleaning it up, in passing.

topiaruss:
Well done. I'm enjoying using it. Thanks!

PR #66, opened 2018-01-29
Zipkin firehose

bplotnick:
This is a quick and dirty implementation of the "firehose mode" (openzipkin/zipkin#1869). The full specification for firehose mode is to come, but the basic idea is that there is a second handler that emits spans regardless of sampling decision.

@msindwan @drolando

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,21 @@ By default, py_zipkin uses a thread local storage for the attributes, which is
 defined in `py_zipkin.stack.ThreadLocalStack`.
 
 
+Firehose mode [EXPERIMENTAL]
+----------------------------
+
+"Firehose mode" records 100% of the spans, regardless of
+sampling rate. This is useful if you want to treat these spans
+differently, e.g. send them to a different backend that has limited
+retention. It works in tandem with normal operation, however there may
+be additional overhead. In order to use this, you add a
+`firehose_handler` just like you add a `transport_handler`.
+
+This feature should be considered experimental and may be removed at
+any time without warning. If you do use this, be sure to send
+asynchronously to avoid excess overhead for every request.
+
+
 License
 -------
 
--- a/py_zipkin/logging_helper.py
+++ b/py_zipkin/logging_helper.py
@@ -48,6 +48,7 @@ class ZipkinLoggingContext(object):
         add_logging_annotation=False,
         client_context=False,
         max_span_batch_size=None,
+        firehose_handler=None,
     ):
         self.zipkin_attrs = zipkin_attrs
         self.thrift_endpoint = thrift_endpoint
@@ -61,6 +62,7 @@ class ZipkinLoggingContext(object):
         self.add_logging_annotation = add_logging_annotation
         self.client_context = client_context
         self.max_span_batch_size = max_span_batch_size
+        self.firehose_handler = firehose_handler
 
     def start(self):
         """Actions to be taken before request is handled.
@@ -87,11 +89,23 @@ class ZipkinLoggingContext(object):
         a success. It also logs the service (`ss` and `sr`) or the client
         ('cs' and 'cr') annotations.
         """
+
+        # FIXME: Should have a single aggregate handler
+        if self.firehose_handler:
+            self._log_spans_with_span_sender(
+                ZipkinBatchSender(self.firehose_handler,
+                                  self.max_span_batch_size)
+            )
+
         if not self.zipkin_attrs.is_sampled:
             return
 
         span_sender = ZipkinBatchSender(self.transport_handler,
                                         self.max_span_batch_size)
+
+        self._log_spans_with_span_sender(span_sender)
+
+    def _log_spans_with_span_sender(self, span_sender):
         with span_sender:
             end_timestamp = time.time()
             # Collect additional annotations from the logging handler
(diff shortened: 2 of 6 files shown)


py_zipkin/logging_helper.py
```diff
@@ -87,11 +89,23 @@ def log_spans(self):
         a success. It also logs the service (`ss` and `sr`) or the client
         ('cs' and 'cr') annotations.
         """
+
+        # FIXME: Should have a single aggregate handler
+        if self.firehose_handler:
+            self._log_spans_with_span_sender(
+                ZipkinBatchSender(self.firehose_handler,
+                                  self.max_span_batch_size)
```
drolando: the 2 max_batch_size don't have to be the same as the zipkin api and collectors don't have the same MTU limitations we have for firehose

py_zipkin/zipkin.py
```diff
@@ -113,7 +113,8 @@ def __init__(
         add_logging_annotation=False,
         report_root_timestamp=False,
         use_128bit_trace_id=False,
-        host=None
+        host=None,
+        firehose_handler=None
```
drolando: Do we want people to re-implement the firehose handler then? I thought we had decided to implement it in py_zipkin directly. In that case this could just be a boolean flag.

py_zipkin/zipkin.py
```diff
@@ -285,7 +288,7 @@ def start(self):
```
drolando: I think you also need to check `self.firehose_handler` in line 246 (self.perform_logging = bool(...)).

github doesn't let me add a comment there since it's far from your changes :(

py_zipkin/zipkin.py
msindwan: Personally, I think we should keep it as a reference to a function. I'd rather see a default implementation that people can import and pass as an argument. That way, users can override the default behaviour.

adriancole:
Thanks for sharing this!

Link didnt work: https://github.com/openzipkin/zipkin/issues/1869 please
link any other details you might have

So assume if doc'd this would be like..

When present, firehose_handler records spans for 100% of requests (ignoring
sample_rate). Implementations might aggregate span metrics or fill a time
limited trace depot. Take care to implement asynchronously to avoid excess
overhead on all requests.

(Probably text isnt awesome)

py_zipkin/logging_helper.py
bplotnick: I added this as a FIXME for this branch, since I am not sure how we want to handle this, so for now we can assume both max sizes are the same.

py_zipkin/zipkin.py
bplotnick: I will add a reference implementation here once we have a design doc and spec for the local firehose daemon. But, at least for now, I agree with @msindwan that we can leave it up to the user.

py_zipkin/zipkin.py
bplotnick: Good call, though I'm not really sure what this code path would even do. In what case would you have no sample rate, no passed in zipkin attrs, and no ongoing trace? Maybe if you called some function that is usually an "inner span"? I dunno. Anyway, I added a test for it and fixed the issue.

py_zipkin/zipkin.py
```diff
@@ -270,6 +274,15 @@ def start(self):
                     flags=existing_zipkin_attrs.flags,
                     is_sampled=existing_zipkin_attrs.is_sampled,
                 )
+            elif self.firehose_handler is not None:
+                # If it has gotten here, the only thing that is
+                # causing a trace is the firehose. So we force a trace
+                # with sample rate of 0
+                report_root_timestamp = True
+                self.zipkin_attrs = create_attrs_for_span(
+                    sample_rate=0.0,
```
msindwan: I'm a little confused here - wouldn't we want a 100% sampling rate on the span? Or is that not how it works?

bplotnick:
@drolando @msindwan I had to pull master. I added documentation in the docstring and the README and parametrized one of the tests that was duplicated. Can you take another look?

py_zipkin/zipkin.py
bplotnick: Sorry I forgot to respond to you. As far as the attrs is concerned, the sample rate is 0%. The firehose is separate from the sample rate.

README.md
```diff
@@ -173,6 +173,21 @@ By default, py_zipkin uses a thread local storage for the attributes, which is
...
+Firehose mode [EXPERIMENTAL]
+----------------------------
+
+"Firehose mode" records 100% of the spans, regardless of
+sampling rate. This is useful if you want to treat these spans
+differently, e.g. send them to a different backend that has limited
+retention. It works in tandem with normal operation, however there may
+be additional overhead. In order to use this, you add a
+`firehose_handler` just like you add a `transport_handler`.
+
```
msindwan: If someone wanted to enable only firehose mode and disable the normal transport, they could simply set the sampling rate to 0% and provide the `firehose_handler`, correct? Could be worth mentioning that here.

bplotnick: I feel like this is implied in "regardless of sampling rate". I'm gonna merge as-is. We'll need to add additional docs about firehose anyway (e.g. example handler and proxy daemon) so if this still isn't clear I'll fix it up later

PR #67, opened 2018-01-30
Added parameter for explicit context

cklein:
A parameter `context_stack` is added to every function of the call stack.
If not provided (default value is `None`), py_zipkin will use the thread local storage
and work as expected.

### DIFF
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -8,15 +8,13 @@ from py_zipkin.exception import ZipkinError
 from py_zipkin.logging_helper import zipkin_logger
 from py_zipkin.logging_helper import ZipkinLoggerHandler
 from py_zipkin.logging_helper import ZipkinLoggingContext
-from py_zipkin.thread_local import get_zipkin_attrs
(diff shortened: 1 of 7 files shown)


py_zipkin/zipkin.py
```diff
@@ -163,6 +164,10 @@ def __init__(
         :param host: Contains the ipv4 value of the host. The ipv4 value isn't
             automatically determined in a docker environment
         :type host: string
+        # CK45
```
bplotnick: What's this comment?

cklein: It was kind of a bookmark for me, forgot to remove it.

cklein:
Hello @bplotnick,
sorry for the delay, I should have answered yesterday, since py_zipkin (as well as distributed_context) kind of blocks the whole project.
I changed my code according to how I understood your comments.
If there are further issues, I suggest that we can address them on IRC (or similar), to reduce the RTT.

bplotnick:
This looks good to me. For future reference, try not to rebase on GH since it messes up the interdiff, and makes it hard to see what has changed 😐

PR #69, opened 2018-02-20
Don't overwrite passed in annotations

bplotnick:
Currently, if you pass in an annotation like `cr` or `sr`, then it will be overwritten when the span closes. The only time you are really going to do this is when you create spans after the fact instead of using the normal start/stop or context manager API.

Internally, we do this and use the `zipkin_logger.debug` API. But since we're trying to get rid of that API, we need to support this. Ideally, this isn't needed and we can actually just switch to the start/stop API, but for now this will do.

### DIFF
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -398,13 +398,12 @@ class zipkin_span(object):
             'ss': end_timestamp,
             'cr': end_timestamp,
         }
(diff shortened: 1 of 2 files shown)


drolando:
@bplotnick anything blocking this PR? Can we merge it?

bplotnick:
@drolando ummm. Nope. I just forgot about this PR!

PR #74, opened 2018-05-04
Add FirehoseHandler from yelp_pyramid to py_zipkin

yannlandry:
I moved `class FirehoseHandler`, along with the related constants and tests from `yelp_pyramid` to here, without further changes.

Test coverage is still 100% and `make test` passes. I only had to add `test_cover_magic_methods()` because py2/py3 call different magic methods (either `__nonzero__()` or `__bool__()`) and the coverage was never 100% otherwise.

I've also transformed the `mock.patch` context managers into decorators purely because otherwise the lines were too long (>83 characters). All seems good for now.

For now nothing was removed from `yelp_pyramid`, and the "old" class is still being used.

### DIFF
--- a/py_zipkin/transport/firehose_handler.py
+++ b/py_zipkin/transport/firehose_handler.py
@@ -1,0 +1,70 @@
+# -*- coding: utf-8 -*-
+import socket
+
+import clog
+import staticconf
+
+
+ZIPKIN_CONFIG_NAMESPACE = 'zipkin'
+ZIPKIN_CONFIG_FILE = '/nail/srv/configs/zipkin.yaml'
+ZIPKIN_CONFIG_MIN_RELOAD_INTERVAL = 5
+
+ZIPKIN_FIREHOSE_IP = '169.254.255.254'
+ZIPKIN_FIREHOSE_PORT = 20182
+
+ZIPKIN_DEFAULT_FIREHOSE_ENABLED = False
+
+
+class FirehoseHandler(object):
+    """A configurable transport handler for firehose mode (CEP935)
+
+    This handler implements truth value testing so that it can be enabled,
+    disabled, or removed, while keeping the py_zipkin logic simple.
+
+    When called with a thrift-encoded zipkin message (typically a
+    thrift-encoded list of spans), it will forward this message over UDP to a
+    proxy, which generally runs as a daemon on the local host.
+    """
+
+    def __init__(self):
+        self.zipkin_namespace = staticconf.NamespaceReaders(
+            ZIPKIN_CONFIG_NAMESPACE)
+        self.config_watcher = staticconf.ConfigFacade.load(
+            ZIPKIN_CONFIG_FILE,
+            ZIPKIN_CONFIG_NAMESPACE,
+            staticconf.YamlConfiguration,
+            min_interval=ZIPKIN_CONFIG_MIN_RELOAD_INTERVAL,
+        )
+        self.firehose_socket = socket.socket(socket.AF_INET,  # Internet
+                                             socket.SOCK_DGRAM)  # UDP
+
+    def __bool__(self):
+        # For python 3
+        return self.is_enabled()
+
+    def __nonzero__(self):
+        # For python 2
+        return self.is_enabled()
+
+    def is_enabled(self):
+        self.config_watcher.reload_if_changed()
+        firehose_enabled = self.zipkin_namespace.read_bool(
+            'enable_firehose',
+            default=ZIPKIN_DEFAULT_FIREHOSE_ENABLED,
+        )
+        return firehose_enabled
+
+    def __call__(self, message):
+        try:
+            self.firehose_socket.sendto(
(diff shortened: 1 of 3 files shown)


py_zipkin/transport/firehose_handler.py
```diff
@@ -0,0 +1,70 @@
+# -*- coding: utf-8 -*-
+import socket
+
+import clog
+import staticconf
+
+
+ZIPKIN_CONFIG_NAMESPACE = 'zipkin'
+ZIPKIN_CONFIG_FILE = '/nail/srv/configs/zipkin.yaml'
```
drolando: can we pass this as argument so it's also useful outside of yelp?

py_zipkin/transport/firehose_handler.py
```diff
@@ -0,0 +1,70 @@
+# -*- coding: utf-8 -*-
+import socket
+
+import clog
```
drolando: You can't use clog here. Can we just use the normal logging library?

py_zipkin/transport/firehose_handler.py
```diff
@@ -0,0 +1,70 @@
...
+ZIPKIN_CONFIG_FILE = '/nail/srv/configs/zipkin.yaml'
+ZIPKIN_CONFIG_MIN_RELOAD_INTERVAL = 5
+
+ZIPKIN_FIREHOSE_IP = '169.254.255.254'
+ZIPKIN_FIREHOSE_PORT = 20182
+
+ZIPKIN_DEFAULT_FIREHOSE_ENABLED = False
+
+
+class FirehoseHandler(object):
```
drolando: I'd call both this class and the file `udp_handler` or something like that, since it's not strictly firehose only.

py_zipkin/transport/firehose_handler.py
```diff
@@ -0,0 +1,70 @@
...
+
+import clog
+import staticconf
+
+
+ZIPKIN_CONFIG_NAMESPACE = 'zipkin'
+ZIPKIN_CONFIG_FILE = '/nail/srv/configs/zipkin.yaml'
+ZIPKIN_CONFIG_MIN_RELOAD_INTERVAL = 5
+
+ZIPKIN_FIREHOSE_IP = '169.254.255.254'
```
msindwan: We hardcoded this to make it easier for us, but since we're including it in py_zipkin now, we should make the IP and Port configurable

py_zipkin/transport/firehose_handler.py
bplotnick: +1

py_zipkin/transport/firehose_handler.py
```diff
@@ -0,0 +1,70 @@
...
+ZIPKIN_FIREHOSE_PORT = 20182
+
+ZIPKIN_DEFAULT_FIREHOSE_ENABLED = False
+
+
+class FirehoseHandler(object):
+    """A configurable transport handler for firehose mode (CEP935)
```
bplotnick: Remove reference to CEP and firehose mode

py_zipkin/transport/firehose_handler.py
```diff
@@ -0,0 +1,70 @@
...
+    This handler implements truth value testing so that it can be enabled,
+    disabled, or removed, while keeping the py_zipkin logic simple.
+
+    When called with a thrift-encoded zipkin message (typically a
+    thrift-encoded list of spans), it will forward this message over UDP to a
+    proxy, which generally runs as a daemon on the local host.
+    """
+
+    def __init__(self):
+        self.zipkin_namespace = staticconf.NamespaceReaders(
```
bplotnick: I'd prefer if this did not require staticconf. I don't want to dep PyStaticConfiguration in py_zipkin.

setup.py
```diff
@@ -19,6 +19,7 @@
     install_requires=[
         'six',
         'thriftpy',
+        'yelp-clog>=0.1.7',
```
bplotnick: Strictly speaking, this PR requires staticconf as well, though I'd like it not to.

py_zipkin/transport/firehose_handler.py
drolando: I feel like staticconf is an ok dependency. At least it's opensourced on github and it's maintaned (yelp-clog is not).

Also it's not yelp specific and should work for anyone.

drolando:
I think @bplotnick's suggestion should be pretty easy and probably cover all the issues. You can just get as argument a user-defined function which returns the sampling probability and a logging function.

py_zipkin/transport/firehose_handler.py
bplotnick: It's not yelp-specific, but it is specific to our use-case of disabling and enabling the transport handler, which is definitely not the normal case. If this is brought into a separate repo and/or as an extras_require, then i'm fine with it. Otherwise, I don't want to suddenly make people pull in a useless dependency.

PR #78, opened 2018-05-23
Always add binary annotations even if the request is not sampled

drolando:
`error` annotations are not being added in firehose mode for non-root spans. We should stop checking if the request is sampled or not and just add them anyway.

### DIFF
--- a/tests/zipkin_test.py
+++ b/tests/zipkin_test.py
@@ -721,6 +721,9 @@ def test_add_sa_binary_annotation():
 
 
 def test_adding_sa_binary_annotation_without_sampling():
+    """Even if we're not sampling, we still want to add binary annotations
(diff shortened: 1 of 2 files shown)


tests/zipkin_test.py
```diff
@@ -733,7 +736,32 @@ def test_adding_sa_binary_annotation_without_sampling():
             service_name='test_service',
             host='1.2.3.4',
         )
-        assert context.logging_context is None
```
msindwan: Why are we removing this assertion?

drolando: Because it's testing something else than what the test name says. Whether `logging_context` is independent of `add_sa_binary_annotation`, so asserting it here doesn't give you anything.

PR #79, opened 2018-05-25
Add support for max batch size in bytes

drolando:
Both the firehose transport (UDP) and the normal scribe/kafka loggers enforce a max payload size. The only config option we have right now is the number of messages in each batch, but that doesn't work well in this case since the span size can vary a lot.

This PR introduces 2 main changes:
- the transport handler is now expected to be a class. This allows us to add an extra method `get_max_payload_bytes` which returns the max payload size for this specific transport. Backward compatibility is maintained by checking whether that method exists, so existing transport functions will keep working.
- we now encode each span individually and then concatenate everything in a thrift list at the end. This lets us maintain an accurate guess of the final object size so that we know whether a new span fits or not.

Most of this logic is very thrift specific and will need to be refactored (probably?) once we add support for JSON encoding. However I decided to not do that now since we're not yet sure how things will look like once we add support for multiple encodings and v2 spans.

Finally, I removed all the `from py_zipkin.thrift import` imports since I was always super confused on whether a function was thrift specific or not in the code. Using `thrift.create_span` seems much clearer to me.


drolando:
@bplotnick one question is, what do we do if one span is bigger than the max payload size? Throw an error? Log an error? Drop it silently?

py_zipkin/logging_helper.py
```diff
@@ -332,16 +328,25 @@ def __init__(self, transport_handler, max_portion_size=None):
         self.transport_handler = transport_handler
         self.max_portion_size = max_portion_size or self.MAX_PORTION_SIZE
 
+        if getattr(self.transport_handler, 'get_max_payload_bytes', None):
+            self.max_payload_bytes = self.transport_handler.get_max_payload_bytes()
```
msindwan: Why not have `max_payload_bytes` as an attribute instead of using a function?

py_zipkin/logging_helper.py
```diff
@@ -365,12 +370,20 @@ def add_span(
...
-        if len(self.queue) >= self.max_portion_size:
-            self.flush()
+        encoded_span = thrift.span_to_bytes(thrift_span)
+
+        # If we've already reached the max batch size or the new span doesn't fit in
+        # max_payload_bytes, send what we've collected until now and start a new batch.
+        if (self.max_payload_bytes and self.current_size + len(encoded_span) > self.max_payload_bytes) \
+                or len(self.queue) >= self.max_portion_size:
+            self.flush(self.queue)
+            self._init_queue()
```
msindwan: I'd call `init_queue` in the flush method

py_zipkin/logging_helper.py
```diff
@@ -332,16 +328,25 @@ def __init__(self, transport_handler, max_portion_size=None):
...
 
     def __exit__(self, _exc_type, _exc_value, _exc_traceback):
         if any((_exc_type, _exc_value, _exc_traceback)):
             error = '{0}: {1}'.format(_exc_type.__name__, _exc_value)
             raise ZipkinError(error)
         else:
-            self.flush()
+            self.flush(self.queue)
+
+    def _init_queue(self):
```
msindwan: nit: maybe `_reset_queue` instead?

py_zipkin/logging_helper.py
```diff
@@ -332,16 +328,25 @@ def __init__(self, transport_handler, max_portion_size=None):
...
+        self._init_queue()
         return self
 
     def __exit__(self, _exc_type, _exc_value, _exc_traceback):
         if any((_exc_type, _exc_value, _exc_traceback)):
             error = '{0}: {1}'.format(_exc_type.__name__, _exc_value)
             raise ZipkinError(error)
         else:
-            self.flush()
+            self.flush(self.queue)
```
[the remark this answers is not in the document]
drolando: I don't think there's any particular reason for it. Probably a leftover from when I was testing different options. I'll revert.

py_zipkin/logging_helper.py
drolando: Because we often end up wanting to reload these values at runtime, so if it's already a function it simplifies our life.

Plus I don't see any particular advantage in having it as an attribute apart from saving 1 line of code.

py_zipkin/logging_helper.py
drolando: sure

drolando:
I had to fix a lot of tests since we used to pass `mock.Mock()` as transport_handler. `Mock` returns another `mock.Mock` when you call `get_max_payload_bytes` which is different than None and is not a float, resulting in a runtime error.

py_zipkin/thrift/__init__.py
```diff
@@ -169,17 +169,29 @@ def create_span(
...
+    thrift_span.write(protocol)
 
     return bytes(transport.getvalue())
+
+
+def encode_bytes_list(binary_thrift_obj_list):  # pragma: no cover
+    """
+    Returns TBinaryProtocol encoded Thrift objects.
+
+    :param binary_thrift_obj_list: thrift objects list to encode
```
bplotnick: strictly speaking it is a list of TBinaryProtocol encoded thrift objects, correct?

py_zipkin/logging_helper.py
```diff
@@ -343,6 +344,10 @@ def __exit__(self, _exc_type, _exc_value, _exc_traceback):
         else:
             self.flush()
 
+    def _reset_queue(self):
+        self.queue = []
+        self.current_size = 5  # Thrift list size
```
bplotnick: Can you pull this magic number out into an ALL_CAPS_CONSTANT

Also can you change the comment to say that this is the thrift list header/overhead, not the size of the list itself, and that the size is in bytes 

I'm also a little bit concerned that we are muddling the concerns of thrift encoding by splitting it apart into two stages (the span encoding and the list construction). I wonder if we can think a little about a cleaner way to do this. Maybe something like subclassing the queue and having `__len__` return the adjusted size

py_zipkin/logging_helper.py
```diff
@@ -332,8 +328,13 @@ def __init__(self, transport_handler, max_portion_size=None):
         self.transport_handler = transport_handler
         self.max_portion_size = max_portion_size or self.MAX_PORTION_SIZE
 
+        if getattr(self.transport_handler, 'get_max_payload_bytes', None):
+            self.max_payload_bytes = self.transport_handler.get_max_payload_bytes()
```
bplotnick: I think it's worthwhile to create an interface/abstract type for the transport handler so that the API is a little more clearly defined. That way we can say `isinstance(ITransportHandler)` or whatever instead of duck-typing.

py_zipkin/logging_helper.py
drolando: > Maybe something like subclassing the queue and having __len__ return the adjusted size

That won't help with doing add_span in 2 stages since we need to check whether the size would go over the limit if we added the current span.

It'd only simplify a bit the if, but that doesn't seem a particularly big improvement to me

py_zipkin/logging_helper.py
drolando: good idea!

py_zipkin/logging_helper.py
drolando: I could do https://i.fluffy.cc/GkFcHZddhX11V1C2JPTXPxQ8CjcRrS00.html, but that doesn't seem a great improvement.

Or maybe it is since then the JSONBatchSender could reuse it?

py_zipkin/logging_helper.py
```diff
@@ -365,12 +370,47 @@ def add_span(
...
         if self.transport_handler and len(self.queue) > 0:
-            message = thrift_objs_in_bytes(self.queue)
+
+            message = thrift.encode_bytes_list(self.queue)
             self.transport_handler(message)
-            self.queue = []
+        self._reset_queue()
+
+
+class ITransportHandler(object):
```
bplotnick: For sanitary reasons, maybe pull this out into another file?

py_zipkin/logging_helper.py
```diff
@@ -365,12 +370,47 @@ def add_span(
...
+
+    def get_max_payload_bytes(self):  # pragma: no cover
+        """Returns the maximum payload size for this transport.
+
+        Most transports have a maximum packet size that can be sent. For example,
+        UDP has a 65507 bytes MTU.
+        py_zipkin automatically batches collected spans for performance reasons.
+        The batch size is gonna be the minimum between `get_max_payload_bytes` and
```
bplotnick: nit: s/gonna/going to/

py_zipkin/logging_helper.py
```diff
@@ -365,12 +370,58 @@ def add_span(
...
-        self.queue.append(thrift_span)
-        if len(self.queue) >= self.max_portion_size:
+        encoded_span = thrift.span_to_bytes(thrift_span)
+
+        # If we've already reached the max batch size or the new span doesn't
+        # fit in max_payload_bytes, send what we've collected until now and
+        # start a new batch.
+        if ((self.max_payload_bytes and
+             self.current_size + len(encoded_span) > self.max_payload_bytes) or
+                len(self.queue) >= self.max_portion_size):
```
bplotnick: this formatting is a little hard to read. maybe split this out into independent statements like (use whatever variable names you like):

```python
is_within_size_limit = (self.current_size + len(encoded_span)) > (self.max_payload_bytes or -1)
is_within_portion_limit = len(self.queue) >= self.max_portion_size
if (is_within_size_limit and is_within_portion_limit):
...
```

py_zipkin/logging_helper.py
```diff
@@ -365,12 +370,58 @@ def add_span(
...
+        :returns: max payload size in bytes or None.
+        """
+        raise NotImplementedError('get_max_payload_bytes is not implemented')
+
+    def send(self, payload):  # pragma: no cover
+        """Sends the encoded payload over the transport.
+
+        :argument payload: encoded list of spans.
+        """
+        raise NotImplementedError('_call is not implemented')
```
bplotnick: nit: fix `NotImplementedError` message?

PR #80, opened 2018-06-01
Decouple py_zipkin internals and thrift

drolando:
As a first step of supporting Span v2 and multiple encodings, I'm decoupling our code from thrift objects. Now we pass around python dicts, lists or `encoding.Endpoint` objects which are generic. All thrift encoding is now done inside `ZipkinBatchSender`, which will make it easier to support JSON encoding in a following PR.

## Internal data structures
The internal model resembles the V2 span, since it's easier to go v2 -> v1 than the other way around as v2 has a better field separation. For now I only had to define a custom `Endpoint` nametuple, since annotations and binary_annotations map perfectly to python dicts.

Instead of building the binary_annotations and annotations thrift objects immediately, I now just pass around the raw dicts and encode them in `ZipkinBatchSender`.

### Endpoint
The V1 format repeats the endpoint everywhere, but it's actually always the same `Endpoint`. So now I'm just passing that along and I add it to the annotations when I'm encoding them to thrift.

### sa annotations
In V1 they just get thrown inside the binary_annotations list, in they're represented as `remoteEndpoint`. So instead of adding them to the annotations list and having to extract them afterwards I'm just passing around the sa endpoint. All other fields in an sa annotation are hardcoded, and are specific to v1 and the specific encoding, so there's no need to pass them around.

In the current implementation sa_annotations is a list so I kept it that way even if I think it's wrong. There will be no way to represent 2 remoteEndpoints in v2.

We don't seem to support ca annotations, so I ignored those for now.

------
Extra changes:
- `from logging import NullHandler`: setup.py says that we only support Python 2.7+, so there's no need to keep a fallback for Python 2.6.
- `perform_logging` inside `zipkin_span` is only used a few lines after in the same function, so there's no need to save it as an object attribute.

cc @adriancole


py_zipkin/encoding.py
```diff
@@ -0,0 +1,41 @@
...
+Endpoint = namedtuple(
+    'Endpoint',
+    ['service_name', 'ipv4', 'ipv6', 'port'],
+)
+
+
+def create_endpoint(port=0, service_name='unknown', host=None):
+    if host is None:
+        try:
+            host = socket.gethostbyname(socket.gethostname())
```
adriancole: might be better to not set ipv4 vs passing localhost.

py_zipkin/logging_helper.py
```diff
@@ -48,14 +43,14 @@ def __init__(
...
-        self.thrift_endpoint = thrift_endpoint
+        self.endpoint = endpoint
         self.log_handler = log_handler
         self.span_name = span_name
         self.transport_handler = transport_handler
         self.response_status_code = 0
         self.report_root_timestamp = report_root_timestamp
         self.binary_annotations_dict = binary_annotations or {}
-        self.sa_binary_annotations = []
+        self.sa_endpoints = []
```
adriancole: do you mean to be plural?

drolando: That's something I wasn't sure about. The old code allowed you to set multiple sa binary annotations, so I kept the same behavior for now.

However I don't see why it'd be useful to be able to set more than one "sa". Also I don't know what to do with it in v2, since I can have only 1 remoteEndpoint there...

py_zipkin/encoding.py
drolando: Are you talking about setting `127.0.0.1` in the except or this `socker.gethostname`? Github shows this comment in an ambiguous position

drolando: Are you talking about setting 127.0.0.1 in the except or this socker.gethostname? Github shows this comment in an ambiguous position

drolando:
@adriancole The default `ipv4` value seems to be `0`, while it's `None` for `ipv6`. Is this correct? Will it work in any encoding?

adriancole:
maybe merge them? bearing in mind I assume this is multiple sa for a single
span..  sometimes we don't know the address until later whereas the service
up front. if that is the use case for multiple sa then maybe merging is ok?

adriancole:
ya defaulting an address to localhost/127.0.0.1

for example it is helpful to know if the listen address or target is
localhost. however a default on exception might mislead..

adriancole:
ipv4 0 is for thrift. in json and proto just leave it out

py_zipkin/encoding.py
drolando: Done

drolando:
Updated code following @adriancole's comments.

- Endpoint doesn't default to '127.0.0.1' anymore. If socket.gethostname fails we simply don't set the ip.
- Only 1 SA annotation is allowed. Trying to add more then one now throws an exception. This is a backward incompatible change, but the old implementation was wrong and had an undefined behavior so I don't think it's too bad

I also renamed the new file as `_encoding_helpers.py` so that it's marked as private as it shouldn't be used directly by users.

py_zipkin/zipkin.py
```diff
@@ -463,16 +460,14 @@ def add_sa_binary_annotation(
...
         )
-        sa_binary_annotation = create_binary_annotation(
-            key=zipkin_core.SERVER_ADDR,
-            value=SERVER_ADDR_VAL,
-            annotation_type=zipkin_core.AnnotationType.BOOL,
-            host=sa_endpoint,
-        )
         if not self.logging_context:
-            self.sa_binary_annotations.append(sa_binary_annotation)
+            if self.sa_endpoint is not None:
```
msindwan: On second thought, I think it makes sense to at least handle the case where the IP address isn't set in the first annotation. 

I.e We can throw an error if the IP address is not None while resetting the SA annotation. This shouldn't complicate the logic too much

drolando: My biggest worry with that is that it complicates the API and makes it hard to understand whether you can or cannot reset the SA endpoint.

Also, what would I do if the 2 endpoints have a different service_name or port?

msindwan: Yeah, I guess ideally if we did support this, we would rename the function to `merge_sa_annotations`. Maybe we could deprecate the `add_sa_binary_annotation` function (i.e print a warning) and recommend that they use the new one?

We could throw a similar error for endpoints that differ in service_name or port as well

drolando: That'd mean adding a new function, they'd still need the old one to add the first sa annotation, no?

It's duable, the thing I'm not convinced of is that it's a good enough use case to be worth complicating the logic to support it. It feels to me they can just wait and set the sa annotation once they know the remote address. I don't see why they'd need to set it in 2 steps.

py_zipkin/logging_helper.py
```diff
@@ -359,14 +338,50 @@ def add_span(
...
+
+        # Binary annotations can be set through debug messages or the
+        # set_extra_binary_annotations registry setting.
+        thrift_binary_annotations = thrift.binary_annotation_list_builder(
+            binary_annotations,
+            thrift_endpoint,
+        )
+
+        # Add sa binary annotation
+        if sa_endpoint is not None:
```
bplotnick: maybe pull this out into its own function?

py_zipkin/thrift/__init__.py
```diff
@@ -64,18 +64,22 @@ def create_endpoint(port=0, service_name='unknown', host=None):
     host is determined automatically
```
bplotnick: can you update the docstring here?

py_zipkin/zipkin.py
```diff
@@ -179,23 +178,21 @@ def __init__(
...
 
-        # Spans that log a 'cs' timestamp can additionally record
-        # 'sa' binary annotations that show where the request is going.
-        # This holds a list of 'sa' binary annotations.
-        self.sa_binary_annotations = []
+        self.logging_context = None
+        self.logging_configured = False
+        self.do_pop_attrs = False
+        # Spans that log a 'cs' timestamp can additionally record a
+        # 'sa' binary annotation that show where the request is going.
```
bplotnick: nit: s/show/shows/

py_zipkin/zipkin.py
bplotnick: This doesn't jive with the zipkinCore documentation:
```
* Indicates a server address ("sa") in a span. Most likely, there's only one.
* Multiple addresses are possible when a client is redirected, or fails to a
* different server ip or port.
```

Though I think it would probably be ok to restrict this (with a note in the documentation) if it makes it easier to move to v2

py_zipkin/_encoding_helpers.py
```diff
@@ -0,0 +1,66 @@
+import socket
```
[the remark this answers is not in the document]
drolando: I'm adding more logic in here in the next PR, that's why I renamed it as `_encoding_helpers`: https://github.com/Yelp/py_zipkin/blob/json_encoding/py_zipkin/_encoding_helpers.py

py_zipkin/zipkin.py
drolando: @adriancole how's that represented in V2 spans?

py_zipkin/logging_helper.py
drolando: I don't see much value in it being a separate functions since it's not something we can reuse. Also it'd be a function that changes the content of one of its arguments rather than return the result which I don't like too much.

bplotnick:
🚢

py_zipkin/zipkin.py
adriancole: the zipkinCore docs need to be fixed. multiple attempts in one span is generally a bad idea, which invalidates the need for multiple sa. we shouldn't recommend it (i probably wrote that note during various attempts to reverse engineer rationale ;))

any merging can be done client side. there's no representation of multiple addresses in v2 either local or remote.

PR #81, opened 2018-06-05
Support JSON encoding for V1 Spans

drolando:
As the title says, this PR adds JSON encoding for V1 spans. This branch is based on `decouple_thrift`, so that one will need to be merged first.

### API changes
`with zipkin_span` now accepts a new argument `encoding`. The valid encodings are listed in an enum imported in `py_zipkin/__init__.py` so users can just do `py_zipkin.Encoding.JSON`. `encoding` defaults to `THRIFT` so this change is backward compatible.

### ZipkinBatchSender
The `ZipkinBatchSender` now receives the encoder instance as argument and uses it to encode the single spans and create the final list. All references to `thrift` are now contained in `_encoding_helpers.py`.

### Encoders
We now have 2 new `_V1ThriftEncoder` and `_V1JSONEncoder`. Every encoder respects the simple `IEncoder` interface.

One issue I encountered is that JSON lists don't have a fixed overhead since single elements are also separated by `", "`. That meant having a `fits` function that receives a bunch of params and tells you whether the new span fits or not. I don't like it much since it receives a lot of arguments, but I didn't know how to better write this API.

### DIFF
--- a/py_zipkin/__init__.py
+++ b/py_zipkin/__init__.py
@@ -1,0 +1,3 @@
+from py_zipkin._encoding_helpers import Encoding
+
+_ = Encoding
--- a/py_zipkin/_encoding_helpers.py
+++ b/py_zipkin/_encoding_helpers.py
@@ -1,10 +1,19 @@
+import json
 import socket
 from collections import namedtuple
+from enum import Enum
+
+from py_zipkin import thrift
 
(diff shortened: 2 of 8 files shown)


drolando:
github didn't like that I deleted the `decouple_thrift` branch. Looks like I'll need to create a new PR

PR #82, opened 2018-06-06
Support JSON encoding for V1 Spans

drolando:
As the title says, this PR adds JSON encoding for V1 spans. This branch is based on `decouple_thrift`, so that one will need to be merged first.

### API changes
`with zipkin_span` now accepts a new argument `encoding`. The valid encodings are listed in an enum imported in `py_zipkin/__init__.py` so users can just do `py_zipkin.Encoding.JSON`. `encoding` defaults to `THRIFT` so this change is backward compatible.

### ZipkinBatchSender
The `ZipkinBatchSender` now receives the encoder instance as argument and uses it to encode the single spans and create the final list. All references to `thrift` are now contained in `_encoding_helpers.py`.

### Encoders
We now have 2 new `_V1ThriftEncoder` and `_V1JSONEncoder`. Every encoder respects the simple `IEncoder` interface.

One issue I encountered is that JSON lists don't have a fixed overhead since single elements are also separated by `", "`. That meant having a `fits` function that receives a bunch of params and tells you whether the new span fits or not. I don't like it much since it receives a lot of arguments, but I didn't know how to better write this API.

cc @adriancole

### DIFF
--- a/py_zipkin/_encoding_helpers.py
+++ b/py_zipkin/_encoding_helpers.py
@@ -1,10 +1,19 @@
+import json
 import socket
 from collections import namedtuple
+from enum import Enum
+
+from py_zipkin import thrift
 
 Endpoint = namedtuple(
     'Endpoint',
     ['service_name', 'ipv4', 'ipv6', 'port'],
 )
+
+
+class Encoding(Enum):
+    THRIFT = 1
+    JSON = 2
 
 
 def create_endpoint(port=0, service_name='unknown', host=None):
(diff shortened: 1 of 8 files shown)


py_zipkin/_encoding_helpers.py
```diff
@@ -64,3 +73,203 @@ def copy_endpoint_with_new_service_name(endpoint, new_service_name):
...
+        return encoded_span, len(encoded_span)
+
+    def encode_queue(self, queue):
+        return thrift.encode_bytes_list(queue)
+
+
+class _V1JSONEncoder(IEncoder):
+    """JSON encoder for V1 spans."""
+
+    def fits(self, queue, current_size, max_size, new_span):
```
adriancole: might expose current_count vs the queue itself

py_zipkin/_encoding_helpers.py
```diff
@@ -1,12 +1,21 @@
...
+
+from py_zipkin import thrift
 
 Endpoint = namedtuple(
     'Endpoint',
     ['service_name', 'ipv4', 'ipv6', 'port'],
 )
 
 
+class Encoding(Enum):
```
adriancole: add _V1 suffix here? allows you to give JSON_V2 later.. or if this is more a about json itsefl leave this alone but add a "format" thing later...

py_zipkin/_encoding_helpers.py
```diff
@@ -64,3 +73,203 @@ def copy_endpoint_with_new_service_name(endpoint, new_service_name):
...
+    def _create_v1_endpoint(self, endpoint):
+        """Converts an Endpoint to a v1 endpoint dict.
+
+        :param endpoint: endpoint object to convert.
+        :type endpoint: Endpoint
+        :returns: dict representing a V1 endpoint.
+        """
+        v1_endpoint = {
+            'serviceName': endpoint.service_name,
+            'port': endpoint.port,
```
adriancole: in v1 iirc ipv4 is not optional (we set it to zero).

py_zipkin/_encoding_helpers.py
```diff
@@ -64,3 +73,203 @@ def copy_endpoint_with_new_service_name(endpoint, new_service_name):
...
+        duration_s,
+        endpoint,
+        sa_endpoint,
+    ):
+
+        span = {
+            'traceId': trace_id,
+            'name': span_name,
+            'id': span_id,
+            'debug': False,
```
adriancole: slightly more efficient to leave out the debug key

py_zipkin/_encoding_helpers.py
drolando: I was thinking of keeping the version and the format separated and so add another `Version(Enum)`. But maybe it's better as you say and have them in a single struct, so we don't have to worry about users setting `V1` with `THRIFT` encoding and similar

adriancole:
your call but yeah my thoughts are with you. ps I am in town today ;)

tests/integration/zipkin_integration_test.py
```diff
@@ -13,6 +18,28 @@
 USECS = 1000000
 
 
+Annotation = namedtuple('Annotation', ['host', 'timestamp', 'value'])
+BinaryAnnotation = namedtuple('BinaryAnnotation', ['key', 'value', 'host'])
+V1Span = namedtuple('V1Span', [
```
bplotnick: yeah! this is kind of what i was talking about above

py_zipkin/_encoding_helpers.py
```diff
@@ -64,3 +73,203 @@ def copy_endpoint_with_new_service_name(endpoint, new_service_name):
...
+        :param queue: queue
+        :type queue: list
+        :returns: (int) overhead in bytes of a list object.
+        """
+        raise NotImplementedError()
+
+    def encode_span(self, *argv):
+        raise NotImplementedError()
+
+    def encode_queue(self, queue):
```
bplotnick: Since we don't have type annotations, it's  probably worth a docstring for these methods to more explicitly define the interface

tests/integration/zipkin_integration_test.py
```diff
@@ -45,7 +76,58 @@ def _decode_binary_thrift_objs(obj):
     return spans
 
 
+def _decode_json_v1_span(obj):
```
bplotnick: I don't think it's really necessary right now, but you could potentially use the jsonschema/swagger definition to build the span class and encoder/decoder programmatically.

I think it's likely that we'll end up going down this path for proto3...

tests/integration/zipkin_integration_test.py
```diff
@@ -90,19 +172,24 @@ def test_starting_zipkin_trace_with_128bit_trace_id(
...
+            assert len(span.trace_id) == 32
 
-    check_span(_decode_binary_thrift_obj(mock_logs[0]))
-    check_span(_decode_binary_thrift_obj(mock_firehose_logs[0]))
+    check_span(decode(mock_logs[0], encoding)[0])
+    check_span(decode(mock_firehose_logs[0], encoding)[0])
 
 
-def test_span_inside_trace():
+@pytest.mark.parametrize('encoding', SUPPORTED_ENCODINGS)
```
bplotnick: I'm not sure if you want to go down this route (i'm thinking that it might actually make things more complicated), but it looks like pytest supports test-suite level parametrization: https://docs.pytest.org/en/latest/example/parametrize.html#a-quick-port-of-testscenarios

tests/logging_helper_test.py
```diff
@@ -457,17 +464,20 @@ def test_zipkin_handler_raises_exception_if_ann_and_bann_not_provided(
...
-    mock_encode_bytes_list,
     empty_annotations_dict,
     empty_binary_annotations_dict,
     fake_endpoint,
 ):
     # This test verifies it's possible to add 1 span without throwing errors.
     # It also checks that exiting the ZipkinBatchSender context manager
     # triggers a flush of all the already added spans.
-    sender = logging_helper.ZipkinBatchSender(MockTransportHandler())
+    encoder = MockEncoder(encoded_queue='foobar')
```
bplotnick: should `encoded_queue` be an iterable of some sort? (ok ok, i know a string is technically an iterable)

py_zipkin/_encoding_helpers.py
```diff
@@ -64,3 +73,203 @@ def copy_endpoint_with_new_service_name(endpoint, new_service_name):
...
+
+        i.e. In JSON is 2 since
+
+        :param queue: queue
+        :type queue: list
+        :returns: (int) overhead in bytes of a list object.
+        """
+        raise NotImplementedError()
+
+    def encode_span(self, *argv):
```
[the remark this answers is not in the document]
drolando: I think v1 and v2 parameters will be the same. I used *argv here mostly for laziness since I didn't want to repeat that list of 10 arguments again. :)

I thought about using a Span namedtuple or class, but I was trying to minimize the code changes. Let me see how big of a difference that makes.

drolando:
Having to deal with the logging_handler is a huge pain and I don't want to waste time updating that logic if we're then going to just delete it. So I'll try to get rid of that first

drolando:
Rebased on top of master. The branch is pretty broken after the rebase though

tests/logging_helper_test.py
drolando: That's the value that gets returned as if it were an encoded queue. So it should be either a string or a byte array.

drolando:
Ok, tests are fixed an we're back to 100% coverage.

I got rid of `debug` since we don't support setting it, so it didn't really have sense to pass around an hardcoded `False` value.

adriancole:
in json nothing except trace/span ids are required generally. it is bad
form to do empty ips. we can go back and markup if we are missing required
flags on the api docs.

in thrift ipv4 is implicitly required so initializes to zero. json didn't
have to do this.

py_zipkin/_encoding_helpers.py
```diff
@@ -15,23 +18,36 @@
 _V1Span = namedtuple(
     'V1Span',
     ['trace_id', 'name', 'parent_id', 'id', 'timestamp', 'duration', 'endpoint',
-     'debug', 'annotations', 'binary_annotations', 'sa_endpoint'],
+     'annotations', 'binary_annotations', 'sa_endpoint'],
 )
 
 
 class Kind(Enum):
+    """ Type of Span. """
```
bplotnick: nit: y/psg says docstrings don't have spaces on either side

py_zipkin/_encoding_helpers.py
```diff
@@ -195,3 +208,208 @@ def copy_endpoint_with_new_service_name(endpoint, new_service_name):
...
+    """Creates encoder object for the given encoding.
+
+    :param encoding: desired output encoding protocol.
+    :type encoding: Encoding
+    :return: corresponding IEncoder object
+    :rtype: IEncoder
+    """
+    if encoding == Encoding.V1_THRIFT:
+        return _V1ThriftEncoder()
+    if encoding == Encoding.V1_JSON:
```
bplotnick: nit: `elif`?

py_zipkin/_encoding_helpers.py
```diff
@@ -195,3 +208,208 @@ def copy_endpoint_with_new_service_name(endpoint, new_service_name):
...
+            span.endpoint.ipv6,
+        )
+
+        thrift_annotations = thrift.annotation_list_builder(
+            span.annotations,
+            thrift_endpoint,
+        )
+
+        # Binary annotations can be set through s messages or the
+        # set_extra_binary_annotations registry setting.
```
bplotnick: I think this comment is out of date?

py_zipkin/_encoding_helpers.py
```diff
@@ -195,3 +208,208 @@ def copy_endpoint_with_new_service_name(endpoint, new_service_name):
...
+        """
+        raise NotImplementedError()
+
+    def encode_span(self, span_builder):
+        """Encodes a single span.
+
+        :param span_builder: span_builder object representing the a span.
+        :type span_builder: SpanBuilder
+        :return: encoded span.
+        :rtype: str or bytes
```
bplotnick: I think this returns a tuple of the span and length, right?

py_zipkin/logging_helper.py
```diff
@@ -158,74 +163,29 @@ def __exit__(self, _exc_type, _exc_value, _exc_traceback):
...
+        encoded_span, size = self.encoder.encode_span(internal_span)
 
         # If we've already reached the max batch size or the new span doesn't
         # fit in max_payload_bytes, send what we've collected until now and
         # start a new batch.
         is_over_size_limit = (
             self.max_payload_bytes is not None and
-            self.current_size + len(encoded_span) > self.max_payload_bytes
+            not self.encoder.fits(self.queue, self.current_size,
+                                  self.max_payload_bytes, encoded_span)
```
bplotnick: nit: prefer one argument per line

py_zipkin/logging_helper.py
```diff
@@ -158,74 +163,29 @@ def __exit__(self, _exc_type, _exc_value, _exc_traceback):
...
-            span.trace_id,
-            span.name,
-            thrift_annotations,
-            thrift_binary_annotations,
-            span.timestamp,
-            span.duration,
-        )
-
-        encoded_span = thrift.span_to_bytes(thrift_span)
+        encoded_span, size = self.encoder.encode_span(internal_span)
```
bplotnick: Is there ever going to be a case where `size` here is not `len(encoded_span)`?

If not, i'd say there's no need to return size.

If so, then you can probably pass `size` instead of `encoded_span` into `encoder.fits`. OR you could invert it a bit and say something like `encoder.bytes_left` or something.

Overall, not a huge deal though, so feel free to ignore if you don't agree.

py_zipkin/_encoding_helpers.py
```diff
@@ -195,3 +208,208 @@ def copy_endpoint_with_new_service_name(endpoint, new_service_name):
...
+        :type new_span: str or bytes
+        :return: True if the new span can be added to the list, False otherwise.
+        :rtype: bool
+        """
+        raise NotImplementedError()
+
+    def encode_span(self, span_builder):
+        """Encodes a single span.
+
+        :param span_builder: span_builder object representing the a span.
```
kaisen: "representing the span"

py_zipkin/_encoding_helpers.py
```diff
@@ -195,3 +208,208 @@ def copy_endpoint_with_new_service_name(endpoint, new_service_name):
...
+                'endpoint': json_sa_endpoint,
+            })
+
+        encoded_span = json.dumps(json_span)
+
+        return encoded_span, len(encoded_span)
+
+    def encode_queue(self, queue):
+        """Concatenates the list to a JSON list"""
+        return '[' + ', '.join(queue) + ']'
```
kaisen: i dont think there needs to be a space after every comma?

py_zipkin/logging_helper.py
kaisen: i vote in favor of getting rid of the 'size' return value. i think it adds a bit of extra complexity to encode_span()

py_zipkin/logging_helper.py
kaisen: so we compute size in encode_span() and then make fits() compute len again? :)

py_zipkin/_encoding_helpers.py
drolando: There's a return before this, so it's not really useful. I actually got a low-priority warning when using the elif.

kaisen:
lg2m!

Issue #61, opened 2017-09-05
Should we still support python "logging" span api


bplotnick:
I don't think that there is much use to be able to say `zipkin_logger.debug(some_dict_form_of_span)`. It only seems to make the api more complicated and cause issues like #23. Can we just get rid of it and store the trace context in thread local?

dmitry-prokopchenkov:
Both hands for removing. I think, zipkin_logger really complicates the code.

adriancole:
Less to explain is better, unless it is compelling. Thx for considering
this!

jimlar:
I have issues with the thread safety (gevent) because of this, let me know if I can help in any way

dmitry-prokopchenkov:
@bplotnick A week was passed. No one asked for saving logging api. Let's remove it?

bplotnick:
@dmitry-prokopchenkov Yeah I think we can remove it. I'll start working on this tomorrow.

adriancole:
Heh probably wont know until after removed :)

On 12 Sep 2017 5:11 pm, "dmitry-prokopchenkov" <notifications@github.com>
wrote:

@bplotnick <https://github.com/bplotnick> A week was passed. No one asked
for saving logging api. Let's remove it?

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/Yelp/py_zipkin/issues/61#issuecomment-328876430>, or mute
the thread

signalpillar:
I second what @jimlar said about thread safety issues. 
One handler is used between multiple greenlets (eventlet) for the different traces that change `parent_span_id` of the handler.

I've changed it a bit to work but didn't have time to get rid of logging completely. What is the progress with that ?


```python

modified   py_zipkin/zipkin.py
@@ -3,6 +3,7 @@ import functools
 import random
 import time
 from collections import namedtuple
+import weakref
 
 from py_zipkin.exception import ZipkinError
 from py_zipkin.logging_helper import zipkin_logger
@@ -96,6 +97,8 @@ class zipkin_span(object):
             do_stuff()
     """
 
+    span_id_to_handler = weakref.WeakValueDictionary()
+
     def __init__(
         self,
         service_name,
@@ -113,7 +116,8 @@ class zipkin_span(object):
         use_128bit_trace_id=False,
         host=None,
         context_stack=None,
-        firehose_handler=None
+        firehose_handler=None,
+        span_id_to_handler=None
     ):
[...]

PR #87, opened 2018-06-20
Remove deprecated logging handler

drolando:
The logging handler interface is deprecated and no one is using it anyway. It also complicates the code quite a lot, so let's get rid of it.

`py_zipkin/logging_helper.py` is now way shorter and much more understandable.

### span_store
The main issue that came up is that until now we stored "client" spans in the logging_handler, which won't be possible anymore. We still need a global singleton storage, so I added a second thread local list.

Also since the new store doesn't go out of scope with the root span, we need to wipe it at the end of `stop`.

### [binary_]annotations_by_span_id
The old code allowed you to set extra annotations via the logging_handler. That would only ever happen if we used the `debug` function (which no one uses), so removing it should be fine.

### span_id, trace_id and parent_span_id
The code was doing something weird with parent_span_id. It was saving it in logger_handler and setting them afterwards with some complicated logic. I can't think of a (still existing) use case for that so I just put the parent_span_id and the trace_id directly into the span dict we're creating.

### Stack is now iterable
We don't want to pop elements out of span_store, since both the firehose handler and the transport handler need to iterate through it. So I made Stack iterable, with the drawback that we have to clear it once we're done.

cc @sjaensch. You guys will probably need to pass a custom `span_store_stack` as well if you upgrade your py_zipkin version.

### DIFF
--- a/py_zipkin/logging_helper.py
+++ b/py_zipkin/logging_helper.py
@@ -1,7 +1,5 @@
 # -*- coding: utf-8 -*-
-import logging
 import time
-from collections import defaultdict
 from logging import NullHandler
 
 from py_zipkin import _encoding_helpers
@@ -10,11 +8,6 @@ from py_zipkin.exception import ZipkinError
 from py_zipkin.transport import BaseTransportHandler
 from py_zipkin.util import generate_random_64bit_string
 
-
-null_handler = NullHandler()
-zipkin_logger = logging.getLogger('py_zipkin.logger')
-zipkin_logger.addHandler(null_handler)
-zipkin_logger.setLevel(logging.DEBUG)
 
 LOGGING_END_KEY = 'py_zipkin.logging_end'
 
@@ -32,10 +25,10 @@ class ZipkinLoggingContext(object):
         self,
         zipkin_attrs,
         endpoint,
-        log_handler,
         span_name,
         transport_handler,
         report_root_timestamp,
+        span_store,
         binary_annotations=None,
         add_logging_annotation=False,
         client_context=False,
@@ -44,10 +37,10 @@ class ZipkinLoggingContext(object):
     ):
         self.zipkin_attrs = zipkin_attrs
         self.endpoint = endpoint
-        self.log_handler = log_handler
         self.span_name = span_name
         self.transport_handler = transport_handler
         self.response_status_code = 0
+        self.span_store = span_store
         self.report_root_timestamp = report_root_timestamp
         self.binary_annotations_dict = binary_annotations or {}
         self.add_logging_annotation = add_logging_annotation
@@ -58,23 +51,18 @@ class ZipkinLoggingContext(object):
         self.sa_endpoint = None
 
     def start(self):
-        """Actions to be taken before request is handled.
-        1) Attach `zipkin_logger` to :class:`ZipkinLoggerHandler` object.
-        2) Record the start timestamp.
-        """
-        zipkin_logger.removeHandler(null_handler)
-        zipkin_logger.addHandler(self.log_handler)
+        """Actions to be taken before request is handled."""
+
+        # Record the start timestamp.
         self.start_timestamp = time.time()
         return self
 
     def stop(self):
         """Actions to be taken post request handling.
-        1) Log the service annotations to scribe.
-        2) Detach `zipkin_logger` handler.
         """
+
+        # Log the service annotations to scribe.
         self.log_spans()
-        zipkin_logger.removeHandler(self.log_handler)
-        zipkin_logger.addHandler(null_handler)
 
     def log_spans(self):
         """Main function to log all the annotations stored during the entire
@@ -92,6 +80,7 @@ class ZipkinLoggingContext(object):
             )
 
         if not self.zipkin_attrs.is_sampled:
+            self.span_store.clear()
             return
 
         span_sender = ZipkinBatchSender(self.transport_handler,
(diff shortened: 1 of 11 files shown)


drolando:
This replaces https://github.com/Yelp/py_zipkin/pull/75

drolando:
FYI @signalpillar, @jimlar and @dmitry-prokopchenkov 

This fixes https://github.com/Yelp/py_zipkin/issues/61

goakley:
> The code was doing something weird with parent_span_id. It was saving it in logger_handler and setting them afterwards with some complicated logic.

Jumping in here, this was actually causing issues when running child spans in a multi-threaded environment, since the `logging_handler.parent_span_id` could be changed by any thread when exiting a span.  This refactor fixes that issue by passing along the `parent_span_id`.  Much obliged and hope to see this merged soon!

py_zipkin/zipkin.py
```diff
@@ -4,10 +4,9 @@
 import time
 from collections import namedtuple
 
+from py_zipkin import thread_local
```
bplotnick: Seems a little weird to import `thread_local` here. What if you were to just create a new (sub?)class of `ThreadLocalStack` that is specific to the span storage? We could also have a subclass specific to zipkin attrs, which I think would simplify the defaulting logic as well

py_zipkin/zipkin.py
```diff
@@ -402,14 +382,16 @@ def stop(self, _exc_type=None, _exc_value=None, _exc_traceback=None):
...
-        )
+        self._span_store.push({
+            'trace_id': self.zipkin_attrs.trace_id,
+            'span_name': self.span_name,
+            'service_name': self.service_name,
+            'parent_span_id': self.zipkin_attrs.parent_span_id,
+            'span_id': self.zipkin_attrs.span_id,
+            'annotations': self.annotations,
+            'binary_annotations': self.binary_annotations,
+            'sa_endpoint': self.sa_endpoint,
```
bplotnick: I think this is the same comment as in the other PR, a span type would be cleaner i think

py_zipkin/stack.py
```diff
@@ -23,6 +23,23 @@ def get(self):
...
+    def __iter__(self):
+        self.index = 0
+        return self
+
+    def __next__(self):  # pragma: no cover
+        return self.next()
+
+    def next(self):
+        if len(self._storage) <= self.index:
+            raise StopIteration
```
bplotnick: I think it is more canonical to have generators consume the items as they iterate. The advantage of having the storage consume the items is that you won't have to call `clear` at the end.

That of course would be O(N) for the default thread local attrs, since its a list (array). But if we change the thread local attrs to use a `deque`, we could do this in constant time (we'd want to make sure to communicate this access pattern in the docstrings so that the implementors of span storage can be sure to use a deque/linked-list).

Another alternative would be to have a method to return an iterator over the items and to not have the storage itself be iterable.

py_zipkin/thread_local.py
```diff
@@ -16,6 +16,17 @@ def get_thread_local_zipkin_attrs():
     return _thread_local.zipkin_attrs
 
 
+def get_thread_local_span_store():
+    """A wrapper to return _thread_local.span_store
```
bplotnick: Would be great to elaborate on the difference between this and the other storage. This one being for output, the other for being for intra-process context propagation

py_zipkin/zipkin.py
drolando: I agree. I just want to keep the 2 PR separated, so I'll add the Span type in the other one.

py_zipkin/stack.py
drolando: We can't have the generator consume the items since we have to iterate twice. Once for the transport_handler and once for the firehose_handler.

bplotnick: Oh right. My concern is that by manually clearing the storage, we might add a new code path in the future and forget to clear the span store. Is there another way that we can ensure that it gets cleared?

py_zipkin/zipkin.py
```diff
@@ -112,6 +111,7 @@ def __init__(
         use_128bit_trace_id=False,
         host=None,
         context_stack=None,
+        span_store_stack=None,
```
bplotnick: reading this again, i'd like to bikeshed the name a bit. There's no reason this has to be a stack, so maybe we can call it something different?

py_zipkin/logging_helper.py
```diff
@@ -58,23 +49,18 @@ def __init__(
...
         self.start_timestamp = time.time()
         return self
 
     def stop(self):
         """Actions to be taken post request handling.
-        1) Log the service annotations to scribe.
-        2) Detach `zipkin_logger` handler.
         """
+
+        # Log the service annotations to scribe.
```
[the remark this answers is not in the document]
drolando: Removed it. I also renamed the function in `emit_spans`, since we cannot know what the transport handler will do. Maybe instead of logging the spans it'll use the HTTP api.

py_zipkin/stack.py
drolando: I integrated some of these comments. I now have `SpanStore` which is basically a deque. 
Deques are iterable, so I don't have to explicitly write the iterator logic. They also already have a `clear` method to wipe their content.

I haven't found a way to avoid calling `span_storage.clear()` unless we want to do a deepcopy of it before calling `emit_spans`.

py_zipkin/zipkin.py
```diff
@@ -165,6 +165,9 @@ def __init__(
         :param context_stack: explicit context stack for storing
             zipkin attributes
         :type context_stack: object
+        :param span_storage: explicit Span storage for storing zipkin spans
+            before they're emitted.
+        : type span_storage: py_zipkin.storage.SpanStorage
```
bplotnick: nit: extra space here

drolando:
Tested this in yelp-main. Everything seems to work fine, I'll only need to update a place that was importing `py_zipkin.stack`

PR #91, opened 2018-07-05
Internal span refactoring

drolando:
Another refactoring split off https://github.com/Yelp/py_zipkin/pull/82 since that PR is too big for me to understand if I'm breaking something.

This change is backward compatible and only refactors py_zipkin internals so that we now send around a `Span` object rather than `dicts`.

It also introduces `kind`, borrowed from the v2 format, that supersededs the old `include` argument. `include` is still supported to maintain backward compatibility and is transparently converted to the right `kind`.

----

Most integration tests pass without changes which is good. The only thing that is different is that if you set the `ss`, `sr`, `cs` or `cr` annotations that won't affect a span timestamp or duration. The annotations themselves are preserved though.

Passing in annotations to force a specific timestamp and duration is weird and not portable (what do you do if you're using v2? Pass the old annotations?). If that's something that people need and use, I can just add 2 extra arguments to `zipkin_span` and allow people to set timestamp and duration there.

### DIFF
--- a/py_zipkin/_encoding_helpers.py
+++ b/py_zipkin/_encoding_helpers.py
@@ -1,10 +1,141 @@
 import socket
 from collections import namedtuple
 
+from enum import Enum
+
+from py_zipkin.exception import ZipkinError
+
+
 Endpoint = namedtuple(
     'Endpoint',
     ['service_name', 'ipv4', 'ipv6', 'port'],
 )
+
+
+_V1Span = namedtuple(
+    'V1Span',
+    ['trace_id', 'name', 'parent_id', 'id', 'timestamp', 'duration', 'endpoint',
+     'debug', 'annotations', 'binary_annotations', 'sa_endpoint'],
+)
+
+
+class Kind(Enum):
+    CLIENT = 'CLIENT'
+    SERVER = 'SERVER'
+    LOCAL = None
+
+
+_DROP_ANNOTATIONS_BY_KIND = {
+    Kind.CLIENT: {'ss', 'sr'},
+    Kind.SERVER: {'cs', 'cr'},
+}
+
+
+class SpanBuilder(object):
+    def __init__(
+        self,
+        trace_id,
+        name,
+        parent_id,
+        span_id,
+        timestamp,
+        duration,
+        annotations,
+        tags,
+        kind,
+        local_endpoint=None,
+        service_name=None,
+        sa_endpoint=None,
+        report_timestamp=True,
+    ):
+        """Internal Span representation. It can generate both v1 and v2 spans.
+
+        It doesn't exactly map to either V1 or V2, since an intermediate format
+        makes it easier to convert to either format.
+
+        :param trace_id: Trace id.
+        :type trace_id: str
+        :param name: Name of the span.
+        :type name: str
+        :param parent_id: Parent span id.
+        :type parent_id: str
+        :param span_id: Span id.
+        :type span_id: str
+        :param timestamp: start timestamp in seconds.
+        :type timestamp: float
+        :param duration: span duration in seconds.
+        :type duration: float
+        :param annotations: Optional dict of str -> timestamp annotations.
+        :type annotations: dict
+        :param tags: Optional dict of str -> str span tags.
+        :type tags: dict
+        :param kind: Span type (client, server, local, etc...)
+        :type kind: Kind
+        :param local_endpoint: The host that recorded this span.
+        :type local_endpoint: Endpoint
+        :param service_name: The name of the called service
+        :type service_name: str
+        :param sa_endpoint: Remote server in client spans.
+        :type sa_endpoint: Endpoint
+        :param report_timestamp: Whether the span should report
+            timestamp and duration.
+        :type report_timestamp: bool
+        """
+        self.trace_id = trace_id
+        self.name = name
+        self.parent_id = parent_id
+        self.span_id = span_id
+        self.kind = kind
+        self.timestamp = timestamp
+        self.duration = duration
+        self.annotations = annotations
+        self.tags = tags
+        self.local_endpoint = local_endpoint
+        self.service_name = service_name
+        self.sa_endpoint = sa_endpoint
+        self.report_timestamp = report_timestamp
+
+        if kind is not None and not isinstance(kind, Kind):
+            raise ZipkinError(
+                'Invalid kind value {}. Must be of type Kind.'.format(kind))
+
+    def build_v1_span(self):
+        """Converts the current span to a V1 Span.
+
+        :returns: newly generated _V1Span
+        """
+        # We are simulating a full two-part span locally, so set cs=sr and ss=cr
+        full_annotations = {
+            'cs': self.timestamp,
+            'sr': self.timestamp,
+            'ss': self.timestamp + self.duration,
+            'cr': self.timestamp + self.duration,
+        }
+
+        if self.kind != Kind.LOCAL:
+            # If kind is not LOCAL, then we only want client or
+            # server side annotations.
+            for ann in _DROP_ANNOTATIONS_BY_KIND[self.kind]:
+                del full_annotations[ann]
+
+        # Add user-defined annotations. We write them in full_annotations
+        # instead of the opposite so that user annotations will override
+        # any automatically generated annotation.
+        full_annotations.update(self.annotations)
+
+        return _V1Span(
+            trace_id=self.trace_id,
+            name=self.name,
+            parent_id=self.parent_id,
+            id=self.span_id,
+            timestamp=self.timestamp if self.report_timestamp else None,
+            duration=self.duration if self.report_timestamp else None,
+            endpoint=self.local_endpoint,
+            debug=False,
+            annotations=full_annotations,
+            binary_annotations=self.tags,
+            sa_endpoint=self.sa_endpoint,
+        )
 
 
 def create_endpoint(port=0, service_name='unknown', host=None):
--- a/py_zipkin/logging_helper.py
+++ b/py_zipkin/logging_helper.py
@@ -1,8 +1,11 @@
 # -*- coding: utf-8 -*-
+import os
 import time
 
-from py_zipkin import _encoding_helpers
 from py_zipkin import thrift
+from py_zipkin import Kind
+from py_zipkin._encoding_helpers import SpanBuilder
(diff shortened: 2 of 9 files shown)


tests/zipkin_test.py
```diff
@@ -228,6 +233,24 @@ def test_zipkin_invalid_include():
...
+def test_zipkin_invalid_kind():
+    with pytest.raises(ZipkinError):
+        with zipkin.zipkin_span(
+            service_name='some_service_name',
+            span_name='span_name',
+            transport_handler=MockTransportHandler(),
+            sample_rate=100.0,
+        ):
+            with zipkin.zipkin_span(
+                    service_name='nested_service',
```
bplotnick: nit: extra indentation

tests/logging_helper_test.py
```diff
@@ -3,6 +3,8 @@
 
 from py_zipkin import _encoding_helpers
 from py_zipkin import logging_helper
+from py_zipkin._encoding_helpers import Kind
```
bplotnick: This can be imported directly from py_zipkin, right?

tests/integration/zipkin_integration_test.py
```diff
@@ -396,3 +396,123 @@ def check_spans(spans):
...
+            'cs', 'cr', 'py_zipkin.logging_end'}
+
+    assert len(mock_logs) == 1
+    check_spans(_decode_binary_thrift_objs(mock_logs[0]))
+
+
+def test_server_span():
+    mock_transport_handler, mock_logs = mock_logger()
+    with zipkin.zipkin_server_span(
+            service_name='test_service_name',
```
bplotnick: nit: extra indentation

py_zipkin/_encoding_helpers.py
```diff
@@ -1,12 +1,163 @@
...
+from py_zipkin.exception import ZipkinError
+
+
 Endpoint = namedtuple(
     'Endpoint',
     ['service_name', 'ipv4', 'ipv6', 'port'],
 )
 
 
+_V1Span = namedtuple(
```
bplotnick: I wonder if it would make more sense to have this subclass some common span base class?

py_zipkin/_encoding_helpers.py
```diff
@@ -1,12 +1,163 @@
...
+    LOCAL = None
+
+
+_DROP_ANNOTATIONS_BY_KIND = {
+    Kind.CLIENT: {'ss', 'sr'},
+    Kind.SERVER: {'cs', 'cr'},
+}
+
+
+class Span(object):
```
bplotnick: This separation of `Span` and `_V1Span` feels a little weird. Like the storage of the data is in `_V1Span`, but it isn't a subclass of `Span`, nor does it make sense for it to be.

Maybe it feels like a `SpanBuilder` rather than a `Span`. Or maybe it is just the name of the thing that feels strange. I dunno. I can't put my finger on it.

py_zipkin/_encoding_helpers.py
```diff
@@ -1,12 +1,163 @@
...
+        :param local_endpoint: The host that recorded this span.
+        :type local_endpoint: Endpoint
+        :param service_name: The name of the called service
+        :type service_name: str
+        :param sa_endpoint: Remote server in client spans.
+        :type sa_endpoint: Endpoint
+        :param include: Which annotations to include
+        :type include: iterable
+        :param report_timestamp: Whether the span should report
+            timestamp and duration.
```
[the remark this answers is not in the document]
drolando: I generate `cr`, `cs`, `sr`, `ss` annotations inside `Span`, so I need to propagate the timestamp and duration.

Since those annotations are V1 specific I think this is a better alternative.

py_zipkin/_encoding_helpers.py
```diff
@@ -1,12 +1,163 @@
...
+            # If kind is not LOCAL, then we only want client or
+            # server side annotations.
+            for ann in _DROP_ANNOTATIONS_BY_KIND[self.kind]:
+                del full_annotations[ann]
+
+        # Add user-defined annotations. We write them in full_annotations
+        # instead of the opposite so that user annotations will override
+        # any automatically generated annotation.
+        for annotation, timestamp in self.annotations.items():
+            full_annotations[annotation] = timestamp
```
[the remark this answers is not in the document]
drolando: good point

tests/integration/zipkin_integration_test.py
```diff
@@ -175,8 +175,8 @@ def check_spans(spans):
         assert nested_span.binary_annotations[0].key == 'nested_key'
         assert nested_span.binary_annotations[0].value == 'nested_value'
         # Local nested spans report timestamp and duration
-        assert nested_span.timestamp == 100 * USECS
-        assert nested_span.duration == 200 * USECS
+        # assert nested_span.timestamp == 100 * USECS
+        # assert nested_span.duration == 200 * USECS
```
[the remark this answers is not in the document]
drolando: So, that one's still there since it's a "feature" that I'm dropping in this change and I couldn't decide whether it's something useful or not.

Specifically before if you passed in custom `cr` & `cs` annotations, py_zipkin would use those to generate duration and timestamp ignoring the automatically generated ones. In the new version that's no longer true. That means `timestamp` and `duration` won't match with `cr`, `cs` anymore.

The other issue is that setting annotations and assuming that'll change the timestamp and duration is a very weird way imo and won't work in V2. Maybe I should just add 2 optional `duration` and `timestamp` arguments to `zipkin_span` and log a deprecation warning if people set those annotations.

py_zipkin/_encoding_helpers.py
drolando: yeah I didn't know how to model these different spans and went back and forth a few times between namedtuples and classes. I like the idea of SpanBuilder though since that's basically what that class is.

tests/integration/zipkin_integration_test.py
drolando: Fixed

py_zipkin/_encoding_helpers.py
drolando: Created `SpanBuilder` class

drolando:
`include` and setting `cr`/`cs`/`sr`/`ss` are now deprecated but still supported. Added 2 extra arguments `timestamp` and `duration` which can be used to override the autogenerated ones.

drolando:
The python 3.4 build is broken but I don't understand why. pip is failing to install tox

drolando:
This seems to be a pip issue: https://github.com/pypa/setuptools/issues/951. Trying to upgrade pip to the latest version

py_zipkin/logging_helper.py
```diff
@@ -178,45 +160,34 @@ def _reset_queue(self):
...
-        span_name,
-        annotations,
-        binary_annotations,
-        timestamp_s,
-        duration_s,
-        endpoint,
-        sa_endpoint,
-    ):
+    def add_span(self, internal_span):
+        span = internal_span.build_v1_span()
```
bplotnick: nit: maybe call this span_builder?

py_zipkin/zipkin.py
```diff
@@ -182,7 +190,25 @@ def __init__(
...
+            # which is represented by kind = None.
+            if 'client' in include and 'server' not in include:
+                self.kind = Kind.CLIENT
+            elif 'client' not in include and 'server' in include:
+                self.kind = Kind.SERVER
+            else:
+                self.kind = Kind.LOCAL
+            warnings.warn(
+                'The include argument is deprecated. Please use kind.',
+                DeprecationWarning,
```
bplotnick: I know I suggested this and I definitely wanna warn the user, but i wonder if this is going to spam the logs too much (and if that's ok).

I wish we had compile-time warnings...

py_zipkin/zipkin.py
```diff
@@ -193,6 +219,12 @@ def __init__(
         else:
             self._span_storage = storage.default_span_storage()
         self.firehose_handler = firehose_handler
+        if kind:
+            self.kind = kind
+        elif self.kind is None:
+            self.kind = Kind.LOCAL
```
bplotnick: it might be clearer here if you merge this with the ` if kind is None and include:` part above

py_zipkin/zipkin.py
```diff
@@ -201,6 +233,29 @@ def __init__(
         # 'sa' binary annotation that shows where the request is going.
         self.sa_endpoint = None
 
+        # It used to  be possible to override timestamp and duration by passing
+        # in the cs/cr or sr/ss annotations. We want to keep backward compatibilty
+        # for now, so this logic overrides self.timestamp and self.duration in the
+        # same way.
+        # This doesn't fit well with v2 spans since those annotations are gone, so
+        # we also log a deprecation warning.
+        if {'sr', 'ss'}.issubset(set(self.annotations.keys())):
```
bplotnick: IMO this is a bit _too_ clever and makes it harder to read than the more verbose `'sr' in self.annotations and 'ss' in self.annotations`

tests/integration/zipkin_integration_test.py
bplotnick: uh. is it?

adriancole:
one thing very scary to me is Kind.LOCAL

I guess this is because you dont want to leave kind unset? if this has to
remain please make sure it doesn't ever accidentally serialize and put lots
of exclamation around it. presence of kind clarifies the remote endpoint so
LOCAL won't be a kind unless the nature of that would change.

drolando:
> I guess this is because you dont want to leave kind unset?

Yeah, it's easier on my side if kind has to be set to something. Otherwise I wouldn't know if it was `None` because we're not propagating it in some code path or because it should actually be `None`.

> make sure it doesn't ever accidentally serialize

Kind.LOCAL maps to None, so it should correctly serialize to `null` in JSON. I'll make sure of this in the other CR

py_zipkin/zipkin.py
drolando: Done

py_zipkin/logging_helper.py
drolando: Well its job is to add the span to the queue, so the name is right.

bplotnick: hah. that is true! I was actually talking about `internal_span`, not `add_span`, but it was ambiguous.

drolando: Lol, @bplotnick ++. Changed it in https://github.com/Yelp/py_zipkin/pull/82

PR #94, opened 2018-08-10
Fix pypi upload.

drolando:
Only upload to pypi after py27 tests. Since our builds are compatible
with all python versions and generate the exact same file, we should
upload it only once.

What's happening right now is that the first tests to finish upload to
pypi and then all the others try to upload as well and fail, failing the
build.

distributions = "sdist bdist_wheel" means from now on we'll upload both
the source tar.gz and a py2.py3-none-any.whl wheel.


bplotnick:
Thanks!

PR #100, opened 2018-09-26
Add V1 to V2 conversion logic.

drolando:
This logic is very convoluted and the code is not pythonic, however we
tried to stay as close as possible to the openzipkin/zipkin
implementation so that it'd be easier to port over future fixes.

@adriancole This is logic that @msindwan wrote a while ago to convert v1 thrift spans to v2 json. This PR is still a work in progress, but can you have a look at whether the code makes sense.
We didn't really know what a bunch of these edge cases protect against, but decided it was best to just replicate the exact same code as zipkin-server.

Also, I'm adding this here due to the discussion we had last Thursday of providing tools to convert v1 to v2. But I'm also fine if we decide this shouldn't live here

### DIFF
--- a/py_zipkin/encoding/__init__.py
+++ b/py_zipkin/encoding/__init__.py
@@ -1,0 +1,96 @@
+# -*- coding: utf-8 -*-
+import json
+
+import six
+
+from py_zipkin.encoding._types import Encoding
+from py_zipkin.encoding._decoders import get_decoder
+from py_zipkin.encoding._encoders import get_encoder
+from py_zipkin.exception import ZipkinError
+
+_V2_ATTRIBUTES = ["tags", "localEndpoint", "remoteEndpoint", "shared", "kind"]
+
+
+def detect_span_version_and_encoding(message):
+    """Returns the span type and encoding for the message provided.
+
+    The logic in this function is a Python port of
+    https://github.com/openzipkin/zipkin/blob/master/zipkin/src/main/java/zipkin/internal/DetectingSpanDecoder.java
+
+    :param message: span to perform operations on.
+    :type message: byte array
+    :returns: span encoding.
+    :rtype: Encoding
+    """
+    # In case message is sent in as non-bytearray format,
+    # safeguard convert to bytearray before handling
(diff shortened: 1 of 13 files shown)


py_zipkin/utils/_thrift_decoder.py
```diff
@@ -0,0 +1,165 @@
...
+    :returns: value as a hex string
+    """
+    result = bytearray(16)
+    _write_hex_long(result, 0, value)
+    return result.decode("utf8")
+
+
+def _write_hex_long(data, pos, value):
+    """
+    Writes an unsigned long value across a byte array.
```
adriancole: might want to qualify big endian

py_zipkin/utils/convert_to_v2_json.py
```diff
@@ -0,0 +1,337 @@
...
+                    span["annotations"].append({
+                        'timestamp': timestamp,
+                        'value': value,
+                    })
+
+        # When bridging between event and span model, you can end up missing a start annotation
+        if self.cs is not None and self._end_timestamp_reflects_span_duration(self.cr):
+            self.cs = {
+                'value': zipkin_core.CLIENT_SEND,
+                'host': self.cr['host'],
```
adriancole: in v2 model there is no host on an annotation

adriancole: oic nm

adriancole:
thanks for the updates!

>

py_zipkin/encoding/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+
+from py_zipkin.encoding._types import Encoding
+from py_zipkin.encoding._decoders import get_decoder
+from py_zipkin.encoding._encoders import get_encoder
+from py_zipkin.exception import ZipkinError
+
+_V2_ATTRIBUTES = ["tags", "localEndpoint", "remoteEndpoint", "shared", "kind"]
+
+
+def detect_span_version_and_encoding(message):
```
adriancole: your code is neater than mine :D

PR #101, opened 2018-09-28
Allow overriding the span name

drolando:
Allow overriding the span name after the span is initialized.

This will be useful in pyramid_zipkin since pyramid doesn't know which route the request matched until the function wrapped by the context manager completes.

I've made this function very specific. I'm unsure whether I should make it more generic and allow overriding any span property: `service_name`, `port`, `timestamp` and `duration`.

### DIFF
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -509,6 +509,21 @@ class zipkin_span(object):
                 raise ValueError('SA annotation already set.')
             self.logging_context.sa_endpoint = sa_endpoint
 
+    def override_span_name(self, name):
+        """Overrides the current span name.
+
+        This is useful if you don't know the span name yet when you create the
+        zipkin_span object. i.e. pyramid_zipkin doesn't know which route the
+        request matched until the function wrapped by the context manager
+        completes.
+
+        :param name: New span name
+        :type name: str
+        """
+        self.span_name = name
+        if self.logging_context:
+            self.logging_context.span_name = name
+
 
 def _validate_args(kwargs):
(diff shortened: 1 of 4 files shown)


py_zipkin/zipkin.py
```diff
@@ -509,6 +509,21 @@ def add_sa_binary_annotation(
...
+
+        This is useful if you don't know the span name yet when you create the
+        zipkin_span object. i.e. pyramid_zipkin doesn't know which route the
+        request matched until the function wrapped by the context manager
+        completes.
+
+        :param name: New span name
+        :type name: str
+        """
+        self.span_name = name
```
bplotnick: Should we even do anything if the span context is closed? e.g. if you have already called `stop()` on a span object?

drolando: I can throw a `ZipkinError` in that case if you think that's better.

With the current code it'd basically change the span name of a useless object since `Span` would have already been added to thread-local. Same behaviour for `update_binary_annotations()`

Which is why I'm not a big fan of having a context manager object that then adds another object to the storage. Because then the client basically has a pointer to something potentially useless.

bplotnick: I don't think it's right to throw the `ZipkinError`, but i think it might be a source of later issues to mutate the state of an object in this way

I don't feel strongly about this. feel free to ship if you disagree.

drolando: It might mostly be confusing for people since the object pointer they have is still valid, but since the real span has already been generated and enqueued changing that one will have no effect

PR #102, opened 2018-10-02
Add support for JSON V2 spans

drolando:
This is useful to add before #100 so that we have `Kind.V2_JSON` as a supported encoding.

The main part of this change is in `_V2JSONEncoder`, which is able to encode a `Span` to the `JSON V2` format. Since several functions were the same as `_V1JSONEncoder` I refactored them away in a base class.

--------

I also added a new integration test file called `encoding_test.py`. Right now we were trying to run the existing integration tests against all the possible encodings we support, which isn't scalable and make the tests complicated.
Also, most of the py_zipkin code is the same for all encodings, so the only thing we need to test for the encodings are what changes between `_V2JSONEncoder` and `_V1JSONEncoder` and `_V1ThriftEncoder`.

This also gives a better idea of what's exactly coming out of the handlers (at least in the python case) since I can compare dicts rather than have many complicated helper functions that make it difficult to understand what's the real object that is being emitted.

cc @adriancole

### DIFF
--- a/py_zipkin/_encoding_helpers.py
+++ b/py_zipkin/_encoding_helpers.py
@@ -2,6 +2,7 @@
 import json
 import socket
 from collections import namedtuple
+from collections import OrderedDict
 
 from enum import Enum
 
@@ -22,6 +23,14 @@ _V1Span = namedtuple(
 )
 
 
+_V2Span = namedtuple(
+    'V2Span',
+    ['trace_id', 'name', 'parent_id', 'id', 'kind', 'timestamp',
+     'duration', 'debug', 'shared', 'local_endpoint', 'remote_endpoint',
+     'annotations', 'tags'],
+)
+
+
 class Kind(Enum):
     """Type of Span."""
     CLIENT = 'CLIENT'
@@ -33,6 +42,7 @@ class Encoding(Enum):
     """Supported output encodings."""
     V1_THRIFT = 1
     V1_JSON = 2
+    V2_JSON = 3
 
 
 _DROP_ANNOTATIONS_BY_KIND = {
@@ -119,12 +129,12 @@ class SpanBuilder(object):
         :rtype: _V1Span
         """
         # We are simulating a full two-part span locally, so set cs=sr and ss=cr
-        full_annotations = {
-            'cs': self.timestamp,
-            'sr': self.timestamp,
-            'ss': self.timestamp + self.duration,
-            'cr': self.timestamp + self.duration,
-        }
+        full_annotations = OrderedDict([
+            ('cs', self.timestamp),
+            ('sr', self.timestamp),
+            ('ss', self.timestamp + self.duration),
+            ('cr', self.timestamp + self.duration),
+        ])
 
         if self.kind != Kind.LOCAL:
             # If kind is not LOCAL, then we only want client or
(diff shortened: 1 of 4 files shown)


py_zipkin/_encoding_helpers.py
```diff
@@ -150,6 +160,32 @@ def build_v1_span(self):
...
+
+        return _V2Span(
+            trace_id=self.trace_id,
+            name=self.name,
+            parent_id=self.parent_id,
+            id=self.span_id,
+            kind=self.kind,
+            timestamp=self.timestamp if self.report_timestamp else None,
+            duration=self.duration if self.report_timestamp else None,
+            debug=False,
```
adriancole: not sure if this implies serialization, but both of the bools should be omitted if false. Also None values like parent timestamp duration etc should also be ommitted.

py_zipkin/_encoding_helpers.py
```diff
@@ -342,24 +380,32 @@ def fits(self, current_count, current_size, max_size, new_span):
...
 
         :param endpoint: endpoint object to convert.
         :type endpoint: Endpoint
-        :return: dict representing a V1 endpoint.
+        :return: dict representing a JSON endpoint.
         :rtype: dict
         """
-        v1_endpoint = {
+        json_endpoint = {
             'serviceName': endpoint.service_name,
```
adriancole: there's a slight difference in v1 vs v2 here. In v1, serviceName could not be null (rather you'd have to hack with empty string).

v2 any value can be left out, though it is poor form to serialize an endpoint with no value.

for this reason of subtlety, I don't tend to share code between the two as documenting why is weirder than the duplicate code :P

py_zipkin/_encoding_helpers.py
```diff
@@ -342,24 +380,32 @@ def fits(self, current_count, current_size, max_size, new_span):
...
             'port': endpoint.port,
```
adriancole: serializing port zero in either v1 or v2 will bloat json for no good reason.

py_zipkin/_encoding_helpers.py
```diff
@@ -409,6 +455,48 @@ def encode_span(self, span_builder):
...
+class _V2JSONEncoder(_BaseJSONEncoder):
+    """JSON encoder for V2 spans."""
+
+    def encode_span(self, span_builder):
+        """Encodes a single span to JSON."""
+        span = span_builder.build_v2_span()
+
+        json_span = {
+            'traceId': span.trace_id,
+            'name': span.name,
```
adriancole: name is optional in v2, and shouldn't be set unless you actually know it (ex don't serialize "unknown"). In v1 we had to hack with empty string when unknown

py_zipkin/_encoding_helpers.py
```diff
@@ -409,6 +455,48 @@ def encode_span(self, span_builder):
...
+        if span.annotations:
+            json_span['annotations'] = [
+                {
+                    'timestamp': int(timestamp * 1000000),
+                    'value': key,
+                }
+                for key, timestamp in span.annotations.items()
+            ]
+
+        encoded_span = json.dumps(json_span)
```
adriancole: does this part serialize debug or shared when they are true?

drolando: The latest commit added `shared`. We don't support `debug` at all with any encoding for now

py_zipkin/_encoding_helpers.py
drolando: Even in V1 JSON I should use `""`?

> though it is poor form to serialize an endpoint with no value
Is there really such a big advantage in letting users emit partial or incomplete spans? It seems to me it complicates the model with little benefits. 

At least here at Yelp partial spans are not allowed. They'd mess up our kafka consumers and break our splunk dashboards since those wouldn't be able to merge a partial span to the rest.

py_zipkin/_encoding_helpers.py
drolando: Even in V1 JSON I should use `""`?

adriancole:
I dont mean to suggest against partial spans.. rather don't serialize

"localEndpoint": {}

make sense?

adriancole:
only in v1 serialize unknown name or serviceName as empty string
in v2 dont serialize empty name or serviceName

drolando:
@adriancole The last commit should solve all the issues you raised

PR #103, opened 2018-10-06
Fix memory leak

drolando:
Added a new `is_transport_configured` helper to `SpanStorage`. This returns `True` only if a transport handler is setup, which allows client context managers to not generate a span if there's no correctly configured root context.

In py_zipkin < 0.13 we used to use `zipkin_logger` to check for this, but now we don't have that global singleton anymore.

The `test_memory_leak` test failed before (span_storage had 10 items into it at the end of the loop) but passes now.

------

I also added a bunch of comments and rearranged a bit the if/else in `zipkin_span` `start()`. The logic is the same as before, but this is hopefully a bit more clear.
I had a very hard time understanding those multiple if/else since some of them only apply to the root context, others only to client spans.

### DIFF
--- a/py_zipkin/storage.py
+++ b/py_zipkin/storage.py
@@ -45,7 +45,29 @@ class ThreadLocalStack(Stack):
 
 
 class SpanStorage(deque):
-    pass
+    def __init__(self):
+        super(SpanStorage, self).__init__()
+        self._is_transport_configured = False
+
+    def is_transport_configured(self):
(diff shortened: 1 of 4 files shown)


drolando:
Note: yelp_memcache uses the internal variable logging_configured, so we'll need to remove that from there. That was the name of a variable that was never exposed, so I don't consider this a backward incompatible change.

Issue #109, opened 2018-10-25
py_zipkin 0.15 prints lots of `invalid escape sequence` on python3.6


drolando:
```
<unknown>:1: DeprecationWarning: invalid escape sequence \]
<unknown>:1: DeprecationWarning: invalid escape sequence \]
<unknown>:1: DeprecationWarning: invalid escape sequence \.
<unknown>:1: DeprecationWarning: invalid escape sequence \.
<unknown>:1: DeprecationWarning: invalid escape sequence \.
<unknown>:1: DeprecationWarning: invalid escape sequence \.
<unknown>:1: DeprecationWarning: invalid escape sequence \[
<unknown>:1: DeprecationWarning: invalid escape sequence \?
<unknown>:1: DeprecationWarning: invalid escape sequence \?
<unknown>:1: DeprecationWarning: invalid escape sequence \[
<unknown>:1: DeprecationWarning: invalid escape sequence \[
<unknown>:1: DeprecationWarning: invalid escape sequence \[
<unknown>:1: DeprecationWarning: invalid escape sequence \[
<unknown>:1: DeprecationWarning: invalid escape sequence \[
<unknown>:1: DeprecationWarning: invalid escape sequence \[
<unknown>:1: DeprecationWarning: invalid escape sequence \[
<unknown>:1: DeprecationWarning: invalid escape sequence \[
<unknown>:1: DeprecationWarning: invalid escape sequence \?
<unknown>:1: DeprecationWarning: invalid escape sequence \?
<unknown>:1: DeprecationWarning: invalid escape sequence \[
<unknown>:1: DeprecationWarning: invalid escape sequence \[
<unknown>:1: DeprecationWarning: invalid escape sequence \[
<unknown>:1: DeprecationWarning: invalid escape sequence \[
<unknown>:1: DeprecationWarning: invalid escape sequence \[
<unknown>:1: DeprecationWarning: invalid escape sequence \[
<unknown>:1: DeprecationWarning: invalid escape sequence \[
<unknown>:1: DeprecationWarning: invalid escape sequence \[
<unknown>:1: DeprecationWarning: invalid escape sequence \]
<unknown>:1: DeprecationWarning: invalid escape sequence \]
<unknown>:1: DeprecationWarning: invalid escape sequence \.
<unknown>:1: DeprecationWarning: invalid escape sequence \.
<unknown>:1: DeprecationWarning: invalid escape sequence \.
[...]

PR #110, opened 2018-10-25
Change DeprecationWarnings to logging.warning

drolando:
DeprecationWarnings are hidden by default in python >= 2.7, so for them
to be usef we need to enable them. Doing so though enabled
DeprecationWarnings for every library and module you import, which
causes a lot of unrelated DeprecationWarnings to be printed. And you
can't do anything for most of them since they're from some other
library you're using.

Let's change these to logging.warning then and revert
DeprecationWarnings to their hidden default.

### DIFF
--- a/py_zipkin/thread_local.py
+++ b/py_zipkin/thread_local.py
@@ -1,8 +1,10 @@
 # -*- coding: utf-8 -*-
+import logging
 import threading
-import warnings
 
 _thread_local = threading.local()
+
+log = logging.getLogger('py_zipkin.thread_local')
 
 
 def get_thread_local_zipkin_attrs():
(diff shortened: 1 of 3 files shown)


drolando:
This solves #109

py_zipkin/thread_local.py
```diff
@@ -42,10 +44,7 @@ def get_zipkin_attrs():
     :rtype: :class:`zipkin.ZipkinAttrs`
     """
     from py_zipkin.storage import ThreadLocalStack
-    warnings.warn(
-        'Use py_zipkin.stack.ThreadLocalStack().get',
-        DeprecationWarning,
-    )
+    log.warning('Use py_zipkin.stack.ThreadLocalStack().get')
```
bplotnick: Since this is no longer a `DeprecationWarning`, can you explicitly call out the deprecation? Same for the other ones.

drolando: good point

Issue #96, opened 2018-08-21
work unnormally when tracing nested spans


nullscc:
I have encountered a problem when I use py_zipkin which version is 0.13.0. But first, let me show you the sample code I use.

## sample code
first, the helper function for `transport_handler` in a file named `handler.py`
```
# handler.py

def default_handler(encoded_span):
    # for version 0.7
    # body = str.encode('\x0c\x00\x00\x00\x01') + encoded_span

    # for version 0.13
    body = encoded_span

    return requests.post(
        "http://127.0.0.1:9411/api/v1/spans",
        data=body,
        headers={'Content-Type': 'application/x-thrift'},
    )
```

and then, the file `demo.py` which is a flask application offer api to client or browser, which export port 5005
```
# demo.py

import requests
from flask import Flask
from py_zipkin.zipkin import zipkin_span,create_http_headers_for_new_span
import time
from handler import default_handler as http_transport

app = Flask(__name__)

@zipkin_span(service_name='webapp', span_name='do_stuff')
def do_stuff():
    headers = create_http_headers_for_new_span()
    requests.get('http://localhost:6000/service1/', headers=headers)
    return 'OK'

@app.route('/')
def index():
    print('recieve request in demo', int(time.time()))
    with zipkin_span(
        service_name='webapp',
        span_name='index',
        transport_handler=http_transport,
        port=5005,
        sample_rate=100, #0.05, # Value between 0.0 and 100.0
    ) as con:
        do_stuff()
    return 'OK', 200

if __name__=='__main__':
    app.run(host="0.0.0.0",port=5005,debug=True)
```

and last but not least, the file named `service1.py` which is another flask application offer an api for `demo.py` to mock a microservice, which expose port 6000 
```
# service1.py
import requests
from flask import Flask, request
from py_zipkin.zipkin import zipkin_span,create_http_headers_for_new_span, ZipkinAttrs
import time
from handler import default_handler as http_transport

app = Flask(__name__)

@zipkin_span(service_name='service1', span_name='service1_do_stuff')
def do_stuff():
    return 'OK'

@app.route('/service1/')
def index():
    with zipkin_span(
        service_name='service1',
        zipkin_attrs=ZipkinAttrs(
            trace_id=request.headers['X-B3-TraceID'],
            span_id=request.headers['X-B3-SpanID'],
            parent_span_id=request.headers['X-B3-ParentSpanID'],
            flags=request.headers['X-B3-Flags'],
[...]

Issue #95, opened 2018-08-19
README says use "Stack()" but Stack() needs one argument


siedentop:
I am trying to use py-zipkin in a multi-threaded context. 

So, I followed the recommendation of the README. However, when running I get an exception because 'storage.Stack()' requires has the signature `def __init__(self, storage)`.

     context_stack = Stack() 
    TypeError: __init__() takes exactly 2 arguments (1 given)

Could you please update the README to work without workarounds?

drolando:
Stack wants you to pass the storage as argument, so you should be using `Stack([])`. I've always found this confusing, so I'll change this to work similar to `SpanStorage`

drolando:
Fixed by #118

PR #111, opened 2018-10-25
Fix multithreading issues with SpanStorage

drolando:
This should fix both #95 and #96. @unapiedra @nullscc

I finally understood why `ThreadLocalStack` had that weird logic. The problem is that when you use a decorator, `SpanStorage` and `Stack` get created at import time. Then your code probably forks or creates different threads and any further call to `thread_local.get_thread_local_span_storage` will actually return a different object.

The easiest way to fix this is by lazily pulling `span_storage` from `thread_local` every time you need it. That will run in the current thread and get it from the correct thread_local.
This unfortunately means `ThreadLocalSpanStorage` needs to override all methods of `SpanStorage` and forward them to `self._storage`. It's not great but it works and it was the cleanest solution I could find.

------

I added a new integration test for this. Without the changes to `py_zipkin.storage` it was failing since `span_storage` would only contain 1 span (the one from the context manager).
After the `py_zipkin.storage` fix it also contains the span from the context manager.

------

@cklein @sjaensch Is the `storage` argument to `Span` useful at all? Can we just default it to an empty list? I can't tell if I'm missing some other corner case here...

### DIFF
--- a/py_zipkin/storage.py
+++ b/py_zipkin/storage.py
@@ -1,6 +1,10 @@
+import logging
 from collections import deque
 
 from py_zipkin import thread_local
+
+
+log = logging.getLogger('py_zipkin.storage')
 
 
 class Stack(object):
@@ -11,8 +15,12 @@ class Stack(object):
     The latter two return None if the stack is empty.
     """
 
-    def __init__(self, storage):
-        self._storage = storage
+    def __init__(self, storage=None):
+        if storage is not None:
+            log.warning('Passing a storage object to Stack is deprecated.')
+            self._storage = storage
+        else:
+            self._storage = []
 
     def push(self, item):
         self._storage.append(item)
@@ -70,5 +78,42 @@ class SpanStorage(deque):
         self._is_transport_configured = configured
 
 
-def default_span_storage():
-    return thread_local.get_thread_local_span_storage()
+class ThreadLocalSpanStorage(object):
+    """
+    ThreadLocalSpanStorage is wrapper around SpanStorage that stores the
+    SpanStorage object in thread local.
+
+    The thread local storage is accessed lazily in every method call,
+    so the thread that calls the method matters, not the thread that
+    instantiated the class. This is important since for example decorators
+    are created at import time and in a different thread than where the
+    rest of the code runs.
+    Every instance shares the same thread local data.
+
+    To make this work I infortunately need to override every method of
+    SpanStorage and call the real span_storage instance.
+    """
+
+    @property
+    def _storage(self):
+        return thread_local.get_thread_local_span_storage()
+
+    def __iter__(self):
+        # No need to override __next__ as __iter__ returns a pointer to
+        # itself. So python will call __next__ directly on the right object.
+        return self._storage.__iter__()
+
+    def __len__(self):
+        return len(self._storage)
+
+    def append(self, el):
+        return self._storage.append(el)
+
+    def clear(self):
+        return self._storage.clear()
+
+    def is_transport_configured(self):
+        return self._storage.is_transport_configured()
+
+    def set_transport_configured(self, configured):
+        self._storage.set_transport_configured(configured)
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -204,7 +204,7 @@ class zipkin_span(object):
         if span_storage is not None:
             self._span_storage = span_storage
         else:
-            self._span_storage = storage.default_span_storage()
+            self._span_storage = storage.ThreadLocalSpanStorage()
         self.firehose_handler = firehose_handler
         self.kind = self._generate_kind(kind, include)
         self.timestamp = timestamp
@@ -256,7 +256,8 @@ class zipkin_span(object):
         if self.sample_rate is not None and not (0.0 <= self.sample_rate <= 100.0):
             raise ZipkinError('Sample rate must be between 0.0 and 100.0')
 
-        if not isinstance(self._span_storage, storage.SpanStorage):
+        if not (isinstance(self._span_storage, storage.SpanStorage) or
+                isinstance(self._span_storage, storage.ThreadLocalSpanStorage)):
             raise ZipkinError('span_storage should be an instance '
                               'of py_zipkin.storage.SpanStorage')
 
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -12,7 +12,7 @@ from py_zipkin.encoding._helpers import create_endpoint
 from py_zipkin.encoding._helpers import Span
 from py_zipkin.exception import ZipkinError
 from py_zipkin.logging_helper import ZipkinLoggingContext
-from py_zipkin.storage import ThreadLocalStack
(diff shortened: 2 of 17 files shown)


py_zipkin/storage.py
```diff
@@ -70,5 +78,42 @@ def set_transport_configured(self, configured):
         self._is_transport_configured = configured
 
 
-def default_span_storage():
-    return thread_local.get_thread_local_span_storage()
+class ThreadLocalSpanStorage(object):
```
bplotnick: hmm. I wish this was of type `SpanStorage`. What if this subclasses `SpanStorage`, and `SpanStorage` can be a base class that looks kind of like a deque, but with only the methods/properties that we need. That way if people want to create their own subtypes, what they need to implement is clear.

I think this would also allow you to move the implementations to the base class and then the subclass would only really need to implement the `storage` property.

tests/integration/multithreading_test.py
```diff
@@ -0,0 +1,49 @@
...
+        span_name='index',
+        transport_handler=transport,
+        sample_rate=100.0,
+        encoding=Encoding.V2_JSON,
+    ):
+        do_stuff()
+
+
+def test_decorator_works_in_a_new_thread():
+    """The zipkin_span decorator is instanciated in a thread and then run in
```
bplotnick:
```suggestion
    """The zipkin_span decorator is instantiated in a thread and then run in
```

py_zipkin/storage.py
drolando: At first I made this a subclass of `SpanStorage`. The issue then was that it was hard to tell if I had forgotten to override some method. Being a subclass of `object` means that if I forget a method it'll just throw an error.

Not a big deal though now that it's implemented though...

signalpillar: I think @bplotnick raised a very good point about the common type between the implementations. Have we considered to create an abstract class for the span storage? In that case it will force to implement the required methods to make it functional.

py_zipkin/zipkin.py
```diff
@@ -256,7 +256,8 @@ def __init__(
         if self.sample_rate is not None and not (0.0 <= self.sample_rate <= 100.0):
             raise ZipkinError('Sample rate must be between 0.0 and 100.0')
 
-        if not isinstance(self._span_storage, storage.SpanStorage):
+        if not (isinstance(self._span_storage, storage.SpanStorage) or
```
signalpillar: That is another place that shows it would be nice to have a base type for the storage to support other implementations in this check and make the message more relevant.
Will it work if instance of `Stack` is passed?

py_zipkin/storage.py
drolando: @signalpillar Can you share an example of how you implemented it? I've no idea how people would implement their own SpanStorage object and what requirements they have, so I'm unsure on what's the best API.

My original idea was that it should be enough instantiate a normal `SpanStorage`, pass it around between functions or store it in your won temporary storage and then just pass it to every `with zipkin_span`. But maybe that's not how you actually use it / there's a better way.

py_zipkin/zipkin.py
drolando: Yup, I'll change this.

> Will it work if instance of Stack is passed?

Nope, `Stack` is not iterable so it won't work.

sjaensch:
@drolando sorry for the late reply! The reason we went with being able to provide the storage object explicitly instead of just relying on thread local storage is cooperative multitasking environments, e.g. event loops / asyncio. If that's something you don't want to support going forward then yes, we can probably drop the `storage` argument again.

drolando:
@sjaensch Yeah I know that, what I don't know is what's the benefit of passing in a storage object.
Wouldn't it be enough to store the `Stack` itself wherever you want and always pass it around? Making also storage configurable seems overkill

py_zipkin/storage.py
signalpillar: I am sorry @drolando, I missed the point that stack and storage are completely different things. We use them differently and they are not related. I needed to implement different stack. I believe we didn't have storage back then. It is quite new thing.

I still believe that it is better to have storage base class as for the different runtime implementation like in the case with eventlet we will need a different implementation.

Is it correct that our current abstract class should look like the following?

```python
class AbstractSpanStorage(abc.ABC):

    @abc.abstractmethod
    def is_transport_configured(self):
        ...

    @abc.abstractmethod
    def set_transport_configured(self, configured):
        ...

    @abc.abstractmethod
    def __iter__(self):
        ...

    @abc.abstractmethod
    def __len__(self):
        ...

    @abc.abstractmethod
    def append(self, el):
        ...

    @abc.abstractmethod
    def clear(self):
        ...
```

py_zipkin/zipkin.py
signalpillar: Sorry, my comment is misleading - stack is used separately from the storage and is not related here.

signalpillar:
The ability to specify own instance of storage is very valuable.
I had to implement open-tracing support for [nameko framework](https://github.com/nameko/nameko) and found it very useful otherwise I would need to make some hacks probably.

There could be a default behaviour but having the option available helps a lot.

signalpillar:
I was thinking about these 2 concepts - storage and stack. Seems like it is obvious that storage is runtime dependent and it could be responsible for creating the stack object that is storated in the storage.

The interface of storage can be extended to access the stack

```
class AbstractSpanStorage(abc.ABC):
    # ...
    @abc.abstractproperty
    def stack(self) -> AbstractStack:
        ...
```

If we use thread-local to keep the state probably it will create the stack instance bound to that threadlocal. If it is corolocal they are related too.
Probably that will cause library interface change.
Just thinking loud.

drolando:
Sorry for the delay in completing this PR.

@signalpillar I kinda followed your advice of integrating span_storage and context_stack. Forcing people to remember to pass both is suboptimal and can lead to bugs.

Also several comments above suggested to make ThreadLocalSpanStorage a subclass of SpanStorage, but that's not possible as the deque and the `_is_logging_configured` bool need to be stored in another object for this to work. They need to be saved inside the `storage` property for this fix to work and the class exposed to the users is basically just a complicated getter.

I've added yet another class and argument called `LocalStorage` which will hold both the context_stack and the span_storage. To implement your own you just need to subclass it and override the `storage` function (which must return an instance of ZipkinStorage). This is backward compatible with providing `context_stack` and `span_storage` for now, but I'll probably remove those in 1.0.

```python
from py_zipkin.storage import LocalStorage
from py_zipkin.storage import ZipkinStorage

class MyLocalStorage(LocalStorage):
    def __init__(self, *argv, **kwargs):
         super(MyLocalStorage, self).__init__(*argv, **kwargs)
         self._storage = ZipkinStorage()

    def storage(self):
         # ofc _storage could be stored anywhere i.e. in gevent.local
         return self._storage
```

`LocalStorage` currently also offers a helper to get the most recent zipkin_attrs `get_zipkin_attrs` and one to check if logging_is_configured.

Names could be better but I didn't know what else to use. I'm open to suggestions.
Also I haven't updated the docs since I have an ongoing branch to rewrite them entirely.

drolando:
#118 has a different implementation for this fix

PR #114, opened 2019-01-24
Rename SpanBuilder to Span and support 'ca' annotation.

drolando:
I'm renaming and exposing py_zipkin.encoding.Span and
py_zipkin.encoding.Endpoint. The next step will be to have `with
zipkin_span` return the Span instance rather than the context manager
itself. That way it'll be possible to update the span fields directly
without need to have setters for all the properties.

```
with zipkin_span(...) as span:
    span.remote_endpoint = create_endpoint(service_name='mysql')
    span.tags.update({'key': 'value'})
```

SpanBuilder was already almost a V2 span, so the changes to it were
minimal.

As a side effect, with this change we also support the 'ca' binary
annotation.

cc @adriancole @amitmokal

### DIFF
--- a/py_zipkin/encoding/_decoders.py
+++ b/py_zipkin/encoding/_decoders.py
@@ -14,7 +14,7 @@ from py_zipkin.encoding._types import Kind
 from py_zipkin.exception import ZipkinError
 from py_zipkin.thrift import zipkin_core
 from py_zipkin.encoding._helpers import Endpoint
-from py_zipkin.encoding._helpers import SpanBuilder
+from py_zipkin.encoding._helpers import Span
 
 _HEX_DIGITS = "0123456789abcdef"
 _DROP_ANNOTATIONS = {'cs', 'sr', 'ss', 'cr'}
@@ -49,8 +49,8 @@ class IDecoder(object):
 
         :param spans: encoded list of spans
         :type spans: bytes
-        :return: list of span builders
-        :rtype: list
+        :return: list of spans
+        :rtype: list of Span
         """
         raise NotImplementedError()
 
@@ -62,10 +62,10 @@ class _V1ThriftDecoder(IDecoder):
 
         :param spans: encoded list of spans
         :type spans: bytes
-        :return: list of span builders
-        :rtype: list
+        :return: list of spans
+        :rtype: list of Span
         """
-        span_builders = []
+        decoded_spans = []
         transport = TMemoryBuffer(spans)
 
         if six.byte2int(spans) == TType.STRUCT:
@@ -76,8 +76,8 @@ class _V1ThriftDecoder(IDecoder):
         for _ in range(size):
             span = zipkin_core.Span()
             span.read(TBinaryProtocol(transport))
-            span_builders.append(self._decode_thrift_span(span))
-        return span_builders
+            decoded_spans.append(self._decode_thrift_span(span))
+        return decoded_spans
 
     def _convert_from_thrift_endpoint(self, thrift_endpoint):
         """Accepts a thrift decoded endpoint and converts it to an Endpoint.
@@ -188,15 +188,14 @@ class _V1ThriftDecoder(IDecoder):
         :param thrift_span: thrift span
         :type thrift_span: thrift Span object
         :returns: span builder representing this span
-        :rtype: SpanBuilder
+        :rtype: Span
         """
         parent_id = None
         local_endpoint = None
         annotations = {}
         tags = {}
         kind = Kind.LOCAL
-        service_name = ''
-        sa_endpoint = None
+        remote_endpoint = None
         timestamp = None
         duration = None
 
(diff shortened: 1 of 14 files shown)


py_zipkin/encoding/_decoders.py
```diff
@@ -220,20 +219,19 @@ def _decode_thrift_span(self, thrift_span):
...
             trace_id=trace_id,
             name=thrift_span.name,
             parent_id=parent_id,
             span_id=self._convert_unsigned_long_to_lower_hex(thrift_span.id),
+            kind=kind,
             timestamp=self.seconds(timestamp or thrift_span.timestamp),
             duration=self.seconds(duration or thrift_span.duration),
+            local_endpoint=local_endpoint,
+            remote_endpoint=remote_endpoint,
+            shared=thrift_span.timestamp is None,
```
adriancole: this is probably incorrect or at least imprecise (checking kind is server can make it more precise unless that is implicit somewhere)

py_zipkin/encoding/_encoders.py
```diff
@@ -95,20 +115,13 @@ def encode_span(self, span_builder):
...
-        # Add sa binary annotation
-        if span.sa_endpoint is not None:
-            thrift_sa_endpoint = thrift.create_endpoint(
-                span.sa_endpoint.port,
-                span.sa_endpoint.service_name,
-                span.sa_endpoint.ipv4,
-                span.sa_endpoint.ipv6,
+        # Add sa/ca binary annotations
+        if v2_span.remote_endpoint:
+            self.encode_remote_endpoint(
```
adriancole: maybe check empty objects or null keys aren't made here. better to send less data

py_zipkin/encoding/_encoders.py
```diff
@@ -176,9 +189,22 @@ def encode_queue(self, queue):
...
 
-    def encode_span(self, span_builder):
+    def encode_remote_endpoint(self, remote_endpoint, kind, binary_annotations):
+        json_remote_endpoint = self._create_json_endpoint(remote_endpoint, True)
+        if kind == Kind.CLIENT:
+            key = 'sa'
+        elif kind == Kind.SERVER:
+            key = 'ca'
+
+        binary_annotations.append({
```
adriancole: if this is v2 json I think the format should be a plain boolean true. in thrift it is type boolean value a single bit (byte with value 1)

py_zipkin/encoding/_helpers.py
```diff
@@ -47,17 +39,17 @@ def __init__(
...
         duration,
-        annotations,
-        tags,
-        kind,
         local_endpoint=None,
-        service_name=None,
-        sa_endpoint=None,
-        report_timestamp=True,
+        remote_endpoint=None,
+        debug=False,
```
adriancole: better to leave these boolean out vs serialize false

py_zipkin/encoding/_encoders.py
drolando: which empty objects or null keys are you referring to?

py_zipkin/encoding/_encoders.py
drolando: In both V1 and V2 the value should be a string according to the spec. This is V1 JSON btw.

py_zipkin/encoding/_helpers.py
drolando: These are just the defaults for the contructor. The decoder drops them if they're not True.

adriancole:
I mean like "localEndpoint":{} or "localEndpoint":{"serviceName": null}

basically scrub for useless space.. usually it is hard to tell if this
happens unless checking string result of encoding

py_zipkin/encoding/_decoders.py
drolando: so check both if timestamp is None and if kind=SERVER?

adriancole:
client and server address are a special case in v1. they aren't a tag. the
special case is that they are key=ca or sa or actually also ma with type
boolean and value true. this horrid complexity is why we made v2

adriancole:
ack thanks

adriancole:
imperfect still because shared is not exactly this (should be set only when
headers are extracted and the span id is reused) however checking kind will
prevent an error where accidentally client is marked as shared which should
never occur

adriancole:
in case helps.. json v1 is almost exactly same as thrift just some slight
encoding shortcuts.
https://github.com/openzipkin/zipkin-api/blob/master/thrift/zipkinCore.thrift#L260

py_zipkin/encoding/_encoders.py
drolando: That's kind of taken care in `py_zipkin/thrift/__init__.py`. However the thrift constructor seems to require all arguments, so it'll depend on what that does. Hopefully it'll just ignore all the `None` arguments.

This is also how it already works right now and given this encoding is deprecated I'm not too interested in optimizing it.

drolando:
Fixed issues an rebased on latest master to fix merge conflicts

PR #118, opened 2019-02-08
Fix multithreading issues

drolando:
This is an alternative implementation for #111.

I didn't particularly like that everything was called *Storage which was
very confusing. Also having to subclass LocalStorage (or SpanStorage)
and override a property felt weird and error prone.
Also the suggested way of getting the current zipkin_attrs by calling
ThreadLocalStack().get() felt weird since why do I need to create a new
object every time?

This PR takes a different approach. Users will need to pass in a
`get_tracer` function, which will be invoked at runtime every time we
need to access the tracer. That makes it work in multithreaded
environments.
Also instead of calling it storage I called the new class Tracer. In the
future I'll probably attach also the transport handlers to it,
completely removing the need for the logging_context.

If users want to access the current context they can just call the
`get_zipkin_attrs()` tracer's function. To access the tracer it's enough
to use `get_default_tracer()` which will return the correct one.

### DIFF
--- a/DEPRECATIONS.rst
+++ b/DEPRECATIONS.rst
@@ -5,21 +5,21 @@ New Tracer interface
 --------------------
 
 The following classes and methods are deprecated and should not be used.
-Please use the Thread interface instead.
+Please use the Tracer interface instead.
 They will be removed in version 1.0.
 
 REASON: they don't work well in multi-threaded environments and can cause
 dropped spans or memory leaks.
 
-- py_zipkin.storage.Stack
-- py_zipkin.storage.ThreadLocalStack
-- py_zipkin.storage.SpanStorage
-- py_zipkin.storage.default_span_storage
-- py_zipkin.thread_local.get_thread_local_zipkin_attrs
-- py_zipkin.thread_local.get_thread_local_span_storage
-- py_zipkin.thread_local.get_zipkin_attrs
-- py_zipkin.thread_local.pop_zipkin_attrs
-- py_zipkin.thread_local.push_zipkin_attrs
+- `py_zipkin.storage.Stack`
+- `py_zipkin.storage.ThreadLocalStack`
+- `py_zipkin.storage.SpanStorage`
+- `py_zipkin.storage.default_span_storage`
+- `py_zipkin.thread_local.get_thread_local_zipkin_attrs`
+- `py_zipkin.thread_local.get_thread_local_span_storage`
+- `py_zipkin.thread_local.get_zipkin_attrs`
+- `py_zipkin.thread_local.pop_zipkin_attrs`
+- `py_zipkin.thread_local.push_zipkin_attrs`
 
 To access the current zipkin_attrs
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/DEPRECATIONS.rst
+++ b/DEPRECATIONS.rst
@@ -43,12 +43,11 @@ your own.
     from py_zipkin import Tracer
 
     tracer = Tracer()
-    with zipkin_span(
+    with tracer.zipkin_span(
         service_name="homepage",
         span_name="get /home",
         transport=MockTransport(),
         sample_rate=100.0,
-        get_tracer=lambda: tracer,
     ):
         # do stuff
 
--- a/DEPRECATIONS.rst
+++ b/DEPRECATIONS.rst
@@ -1,0 +1,101 @@
+Deprecations: how to migrate
+============================
+
+New Tracer interface
+--------------------
+
+The following classes and methods are deprecated and should not be used.
+Please use the Thread interface instead.
+They will be removed in version 1.0.
+
+REASON: they don't work well in multi-threaded environments and can cause
+dropped spans or memory leaks.
+
+- py_zipkin.storage.Stack
+- py_zipkin.storage.ThreadLocalStack
+- py_zipkin.storage.SpanStorage
+- py_zipkin.storage.default_span_storage
+- py_zipkin.thread_local.get_thread_local_zipkin_attrs
+- py_zipkin.thread_local.get_thread_local_span_storage
+- py_zipkin.thread_local.get_zipkin_attrs
+- py_zipkin.thread_local.pop_zipkin_attrs
+- py_zipkin.thread_local.push_zipkin_attrs
+
+To access the current zipkin_attrs
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: python
+
+    from py_zipkin import get_default_tracer
+
+    zipkin_attrs = get_default_tracer().get_zipkin_attrs()
+    get_default_tracer().push_zipkin_attrs(zipkin_attrs)
+
+To override the default tracer and provide your own in a multi-threaded env
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can instantiate your own Tracer instance and pass it to ALL your zipkin_span
+context managers and decorators. You'll need to propagate the tracer around on
+your own.
+
+.. code-block:: python
+
+    from py_zipkin import Tracer
+
+    tracer = Tracer()
+    with zipkin_span(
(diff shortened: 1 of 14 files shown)


DEPRECATIONS.rst
```diff
@@ -0,0 +1,101 @@
...
+
+    from py_zipkin import Tracer
+
+    tracer = Tracer()
+    with zipkin_span(
+        service_name="homepage",
+        span_name="get /home",
+        transport=MockTransport(),
+        sample_rate=100.0,
+        get_tracer=lambda: tracer,
```
adriancole: interesting :)

DEPRECATIONS.rst
```diff
@@ -0,0 +1,101 @@
...
+    ):
+        # do stuff
+
+Alternatively you can use the `get_default_tracer` and `set_default_tracer`
+functions to reset the default tracer every time you switch thread or
```
adriancole: we use the term "current span" and "current tracer" in brave in case that is interesting to you

DEPRECATIONS.rst
```diff
@@ -0,0 +1,101 @@
...
+    with zipkin_span(
+        service_name="homepage",
+        span_name="get /home",
+        include=('server',),
+    ):
+        pass
+
+    # New code, uses Kind
+    with zipkin_span(
+        service_name="homepage",
```
adriancole: please don't add examples that add service names per span as people get this wrong constantly, only because the api makes it possible :) service names are almost always tracer scoped. for top-level examples it would be better to not encourage span-scoped service names (as there's no mechanism to propagate that name downstream and even it there was, this would be an advanced topic imho)

drolando: Right now service_name is a required argument in py_zipkin. But I agree with you that we currently have the issue of inconsistent service names and that it'd probably make more sense as a tracer argument

DEPRECATIONS.rst
drolando: Yeah I thought about it and decided against using "current". The issue is that since we support passing in a custom tracer to every context manager there's no way for a generic helper like this to return that. So I think "default" is a better name since it users are not passing in a custom one "default" sound what you're looking for, while if you use a custom tracer then you'll realize this is a different one

adriancole: that's fine was just mentioning in case it hadn't been considered.thx

DEPRECATIONS.rst
```diff
@@ -0,0 +1,101 @@
...
+============================
+
+New Tracer interface
+--------------------
+
+The following classes and methods are deprecated and should not be used.
+Please use the Thread interface instead.
+They will be removed in version 1.0.
+
+REASON: they don't work well in multi-threaded environments and can cause
```
bplotnick: This is really excellent. To have reasoning behind deprecations is something that i think end users will appreciate

DEPRECATIONS.rst
```diff
@@ -0,0 +1,101 @@
+Deprecations: how to migrate
+============================
+
+New Tracer interface
+--------------------
+
+The following classes and methods are deprecated and should not be used.
+Please use the Thread interface instead.
```
bplotnick: There's a number of references to `Thread`, which I think are relics of a prior iteration

drolando:
@bplotnick That's an interesting idea and it'd make our api closer to what everyone else is doing. Plus you don't need to write a tracer getter: you just pass around `tracer` somehow and just always reuse it. Or get the default one. Which makes the code a bit simpler.

It's very easy to make this backward compatible so that we wouldn't break everyone's code for now.
```diff
diff --git a/py_zipkin/storage.py b/py_zipkin/storage.py
index 0731ce6..71c9467 100644
--- a/py_zipkin/storage.py
+++ b/py_zipkin/storage.py
@@ -95,6 +95,11 @@ class Tracer(object):
     def is_transport_configured(self):
         return self._is_transport_configured

+    def zipkin_span(self, *argv, **kwargs):
+        from py_zipkin.zipkin import zipkin_span
+        kwargs['_tracer'] = self
+        return zipkin_span(*argv, **kwargs)
+

 class Stack(object):
     """
diff --git a/py_zipkin/zipkin.py b/py_zipkin/zipkin.py
index a4100c6..1dc6cd8 100644
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -116,7 +116,7 @@ class zipkin_span(object):
         timestamp=None,
         duration=None,
         encoding=Encoding.V1_THRIFT,
-        get_tracer=None,
+        _tracer=None,
     ):
         """Logs a zipkin span. If this is the root span, then a zipkin
         trace is started as well.
@@ -187,9 +187,9 @@ class zipkin_span(object):
         :type duration: float
         :param encoding: Output encoding format, defaults to V1 thrift spans.
         :type encoding: Encoding
-        :param get_tracer: getter function that returns the current Tracer. Will be
-            invoked at runtime every time we need to access the tracer object.
-        :type get_tracer: function
+        :param _tracer: Current tracer object. This argument is passed in automatically
+            when you create a zipkin_span from a Tracer.
+        :type _tracer: Tracer
         """
         self.service_name = service_name
         self.span_name = span_name
@@ -211,10 +211,7 @@ class zipkin_span(object):
         self.timestamp = timestamp
         self.duration = duration
         self.encoding = encoding
-        if get_tracer is not None:
-            self.get_tracer = get_tracer
-        else:
-            self.get_tracer = get_default_tracer
+        self._tracer = _tracer

         self._is_local_root_span = False
         self.logging_context = None
@@ -304,11 +301,17 @@ class zipkin_span(object):
                 timestamp=self.timestamp,
                 duration=self.duration,
                 encoding=self.encoding,
-                get_tracer=self.get_tracer,
+                _tracer=self._tracer,
             ):
                 return f(*args, **kwargs)
         return decorated

+    def get_tracer(self):
+        if self._tracer is not None:
+            return self._tracer
+        else:
+            return get_default_tracer()
+
     def __enter__(self):
         return self.start()

diff --git a/tests/integration/multithreading_test.py b/tests/integration/multithreading_test.py
index 563b53f..748ea8c 100644
--- a/tests/integration/multithreading_test.py
+++ b/tests/integration/multithreading_test.py
@@ -3,11 +3,14 @@ import json
 from threading import Thread

 from py_zipkin import Encoding
+from py_zipkin.storage import get_default_tracer
 from py_zipkin.zipkin import zipkin_span
 from tests.test_helpers import MockTransportHandler


@zipkin_span(service_name='service1', span_name='service1_do_stuff')
 def do_stuff():
     return 'OK'

@@ -20,7 +23,7 @@ def run_inside_another_thread(transport):
         need a way to return the results to the main thread.
     :type transport: MockTransportHandler
     """
-    with zipkin_span(
+    with get_default_tracer().zipkin_span(
         service_name='webapp',
         span_name='index',
         transport_handler=transport,
```
-------------------

The only catch is that you can't use this for decorators. There the default tracer is usually the wrong one and you would have the same bug we have right now.

This fails the new multithreading test.
```python
tracer = get_default_tracer()
@tracer.zipkin_span(service_name='service1', span_name='service1_do_stuff')
def fn():
    pass
```

This is syntactically wrong and is basically the same thing as the previous code snippet.
```python
@get_default_tracer().zipkin_span(service_name='service1', span_name='service1_do_stuff')
def fn():
    pass
```

This does the right thing and works. Even with the new code since `get_tracer` defaults to calling `get_default_tracer` at runtime.
```python
@zipkin_span(service_name='service1', span_name='service1_do_stuff')
def fn():
    pass
```
Which means decorators would need to keep using the old interface, while context managers can use the new one.

That's fine for me, but it might be confusing to users... On the other hand if the docs clearly show how to use context managers and decorators with examples it should be fine.

DEPRECATIONS.rst
drolando: Lol I kept writing Thread rather than Tracer everywhere

drolando:
Fixed all references to `Thread` to `Tracer`.

The `Create zipkin_span from tracer` commit also allows creating the zipkin span from the tracer, while keeping backward compatibility by still allowing the current format.
Now there's no need to specify a custom `get_tracer` and the tracer itself can pass a reference to itself when the span is created from it.

bplotnick:
This is really awesome. Great work on this. :shipit:

PR #120, opened 2019-02-21
Make ThreadLocalStack() get context from tracer

drolando:
py-zipkin 0.18.0 broke `ThreadLocalStack().get()` since that's now using a different thread_local variable than the tracer.

The only real change here is to make the thread_local module use the same thread_local variables as the tracer. The rest is just tests updates.

Plus a small bug fix for tox.ini so that it doesn't build the wheel for thriftpy2 since that wheel doesn't get built correctly under pypy and pollutes the local cache.

### DIFF
--- a/py_zipkin/thread_local.py
+++ b/py_zipkin/thread_local.py
@@ -1,8 +1,7 @@
 # -*- coding: utf-8 -*-
 import logging
-import threading
 
-_thread_local = threading.local()
+from py_zipkin.storage import get_default_tracer
 
 log = logging.getLogger('py_zipkin.thread_local')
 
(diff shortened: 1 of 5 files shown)


py_zipkin/thread_local.py
```diff
@@ -43,10 +40,7 @@ def get_thread_local_span_storage():
     """
     log.warning('get_thread_local_span_storage is deprecated. See DEPRECATIONS.rst'
                 ' for details on how to migrate to using Tracer.')
-    if not hasattr(_thread_local, 'span_storage'):
-        from py_zipkin.storage import SpanStorage
-        _thread_local.span_storage = SpanStorage()
-    return _thread_local.span_storage
+    return get_default_tracer()._span_storage
```
amitmokal: How about implementations for:
def get_zipkin_attrs(),
def pop_zipkin_attrs(),
push_zipkin_attrs(zipkin_attr):

They should also work with get_default_tracer()?

amitmokal: Also, why not get rid of 
_thread_local = threading.local() [py_zipkin/thread_local.py].
Since only Tracer will be used and everything should be stored in _thread_local_tracer = threading.local()

drolando: > How about implementations for: get_zipkin_attrs(), pop_zipkin_attrs(), push_zipkin_attrs(zipkin_attr)

They're also being updated since they use `ThreadLocalStack().get()`.

> Also, why not get rid of _thread_local = threading.local()

Yeah I'm already deleting it

PR #121, opened 2019-02-25
Add protobuf encoding

drolando:
This PR adds support for the protobuf v3 encoding.

I tested this by manually generating a few traces, writing them to a
file and sending them to zipkin with curl.

I'm checking in the zipkin.proto file. It's not really necessary but it
seemed to make sense to keep it next to the generated code.

I also changed `encoding.create_endpoint` to accept an `use_defaults`
param which disables guessing default values. That makes tests easier
since the guessed ips would change from host to host.

cc @adriancole for reviewing


py_zipkin/encoding/_encoders.py
```diff
@@ -21,6 +22,8 @@ def get_encoder(encoding):
         return _V1JSONEncoder()
     if encoding == Encoding.V2_JSON:
         return _V2JSONEncoder()
+    if encoding == Encoding.V2_PROTOBUF:
```
adriancole: not necessarily important to you, but we call this PROTO3 to differentiate between the format. There was no V1 encoding.. anyway you can decide what you think is right https://github.com/openzipkin/zipkin/blob/master/zipkin/src/main/java/zipkin2/codec/SpanBytesEncoder.java#L110

py_zipkin/encoding/_encoders.py
```diff
@@ -298,3 +301,32 @@ def encode_span(self, span):
...
         return encoded_span
+
+
+class _V2ProtobufEncoder(IEncoder):
+    """Protobuf encoder for V2 spans."""
+
+    def fits(self, current_count, current_size, max_size, new_span):
+        """Checks if the new span fits in the max payload size.
+
+        We can't encode protobuf spans one by one and just sum their sizes as we
```
adriancole: I think this is actually not true. see https://github.com/openzipkin/zipkin/blob/master/zipkin/src/main/java/zipkin2/codec/Encoding.java#L63

py_zipkin/encoding/protobuf/__init__.py
```diff
@@ -0,0 +1,167 @@
+# -*- coding: utf-8 -*-
+import socket
+import struct
+
+from py_zipkin.encoding._types import Kind
+from py_zipkin.util import unsigned_hex_to_signed_int
+try:
+    from py_zipkin.encoding.protobuf import zipkin_pb2
+except ImportError:  # pragma: no cover
+    zipkin_pb2 = None
```
adriancole: this is the small confusion I perceive. proto3 exists and this makes me think.. is this proto2 encoding?

Confusion may only exist in my head

py_zipkin/encoding/protobuf/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+    if span.parent_id:
+        pb_kwargs['parent_id'] = _hex_to_bytes(span.parent_id)
+
+    pb_kwargs['id'] = _hex_to_bytes(span.span_id)
+
+    pb_kind = _get_protobuf_kind(span.kind)
+    if pb_kind:
+        pb_kwargs['kind'] = pb_kind
+
+    pb_kwargs['name'] = span.name
```
adriancole: these three fields are nullable btw

py_zipkin/encoding/protobuf/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+
+    :param annotations: annotations dict.
+    :type annotations: dict
+    :return: corresponding protobuf's list of annotations.
+    :rtype: list
+    """
+    pb_annotations = []
+    for value, ts in annotations.items():
+        pb_annotations.append(zipkin_pb2.Annotation(
+            timestamp=int(ts * 1000 * 1000),
```
adriancole: heh :P precision not so hot huh?

py_zipkin/encoding/protobuf/zipkin.proto
```diff
@@ -0,0 +1,214 @@
+// Copyright 2018 The OpenZipkin Authors
```
adriancole: maybe cite where this file is from, for maintenance pov?

py_zipkin/encoding/protobuf/__init__.py
drolando: Yeah this is very confusing. https://github.com/grpc/grpc/issues/15444 explains the 2 suffix, which has nothing to do with the protobuf version.

py_zipkin/encoding/protobuf/__init__.py
drolando: Also traceId and id? What'd be the use case to emit a span without them?

py_zipkin/encoding/protobuf/__init__.py
drolando: should be mostly right no? ts is in seconds and a float, so I multiply but 1000*1000 to convert to microseconds and then to int.

py_zipkin/encoding/protobuf/__init__.py
adriancole: traceId and id are the only required fields in all "v2" formats.. though it is silly to send these alone. The schema language would be a sortof AnyOf thing, where you should send at least something else.. can't really express this.

py_zipkin/encoding/protobuf/__init__.py
adriancole: ah didn't know it was a float.

py_zipkin/encoding/protobuf/__init__.py
drolando: Oh you meant name, timestamp and duration.

Yeah, I'll add the check but there's currently no way in py_zipkin to create a span that doesn't have them.

py_zipkin/encoding/protobuf/zipkin.proto
drolando: The makefile just pulls it from zipkin-api so I'd rather not modify it at all: https://github.com/drolando/py_zipkin/blob/add_protobuf_encoding/py_zipkin/encoding/protobuf/Makefile#L2

adriancole:
> The makefile just pulls it from zipkin-api so I'd rather not modify it at all: https://github.com/drolando/py_zipkin/blob/add_protobuf_encoding/py_zipkin/encoding/protobuf/Makefile#L2
ah ok

py_zipkin/encoding/_encoders.py
drolando: Before I was trying to convert each span to bytes and concatenate them which didn't work. Then I realized that if I encoded them as 1 element ListOfSpans I could very easily concatenate everything

PR #123, opened 2019-03-25
log error over raise ZipkinError

amitmokal:
A ZipkinError was being raised to due to a bad span id (multiple span ids) from the service layer. Services do not explicitly handle ZipkinError. And do not want their normal functionality to fail. Although tracing will not work (emit spans), we log an error in the (py_zipkin) library.

Also in case of an error, the tracer needs to be cleaned up. Otherwise the context gets carried over.

### DIFF
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -496,8 +496,7 @@ class zipkin_span(object):
             try:
                 self.logging_context.stop()
             except Exception as ex:
-                err_msg = 'Error emitting zipkin trace, service:{}, {}'.format(
-                    self.service_name,
+                err_msg = 'Error emitting zipkin trace. {}'.format(
                     repr(ex),
                 )
                 log.error(err_msg)
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -495,14 +495,15 @@ class zipkin_span(object):
         if self.logging_context:
             try:
                 self.logging_context.stop()
-            except Exception:
-                self.get_tracer().clear()
-                err_msg = 'Error while emitting zipkin trace. {}'.format(
+            except Exception as ex:
+                err_msg = 'Error emitting zipkin trace, service:{}, {}'.format(
                     self.service_name,
+                    repr(ex),
                 )
                 log.error(err_msg)
             finally:
                 self.logging_context = None
+                self.get_tracer().clear()
                 self.get_tracer().set_transport_configured(configured=False)
                 return
 
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -34,8 +34,6 @@ ZipkinAttrs = namedtuple(
 
 ERROR_KEY = 'error'
 
-log = logging.getLogger('py_zipkin.zipkin')
-
 
 class zipkin_span(object):
     """Context manager/decorator for all of your zipkin tracing needs.
@@ -466,11 +464,7 @@ class zipkin_span(object):
         return self
 
     def __exit__(self, _exc_type, _exc_value, _exc_traceback):
-        try:
-            self.stop(_exc_type, _exc_value, _exc_traceback)
-        except Exception:
-            log.error('ZipkinError', exc_info=True)
-            self.get_tracer().clear()
+        self.stop(_exc_type, _exc_value, _exc_traceback)
 
     def stop(self, _exc_type=None, _exc_value=None, _exc_traceback=None):
         """Exit the span context. Zipkin attrs are pushed onto the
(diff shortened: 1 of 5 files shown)


py_zipkin/logging_helper.py
```diff
@@ -156,7 +160,8 @@ def __exit__(self, _exc_type, _exc_value, _exc_traceback):
                 _exc_type.__name__,
                 _exc_value,
             )
-            raise ZipkinError(error)
+            self.tracer.clear()
+            logger.error('ZipkinError', json.dumps(error))
```
drolando: error is already a string, no need to do json.dumps()

py_zipkin/logging_helper.py
```diff
@@ -156,7 +160,8 @@ def __exit__(self, _exc_type, _exc_value, _exc_traceback):
                 _exc_type.__name__,
                 _exc_value,
             )
-            raise ZipkinError(error)
+            self.tracer.clear()
```
drolando: This is taken care in `zipkin_span`, no? Now we don't throw an exception anymore, so we'll always go through the cleanup logic

drolando: Can you add an integration test to verify that?

py_zipkin/logging_helper.py
```diff
@@ -89,6 +91,7 @@ def emit_spans(self):
             return
 
         span_sender = ZipkinBatchSender(self.transport_handler,
+                                        self._get_tracer(),
```
drolando: Can you pass the getter function for consistency? It's probably fine to pass the tracer object itself here, but that'd break if we will ever make ZipkinBatchSender asynchronous.

py_zipkin/zipkin.py
```diff
@@ -465,7 +466,11 @@ def start(self):
         return self
 
     def __exit__(self, _exc_type, _exc_value, _exc_traceback):
-        self.stop(_exc_type, _exc_value, _exc_traceback)
+        try:
+            self.stop(_exc_type, _exc_value, _exc_traceback)
+        except Exception:
+            log.error('ZipkinError', exc_info=True)
```
drolando: what does this actually output? "ZipkinError" is super vague

py_zipkin/zipkin.py
```diff
@@ -465,7 +466,11 @@ def start(self):
         return self
 
     def __exit__(self, _exc_type, _exc_value, _exc_traceback):
-        self.stop(_exc_type, _exc_value, _exc_traceback)
+        try:
```
drolando: This won't work for users calling `.stop()` explicitly

amitmokal:
Moved the exception handling inside of stop()
The integration test remains the same as before
An unit test introduced for code coverage inside the exception block

py_zipkin/zipkin.py
```diff
@@ -494,10 +493,18 @@ def stop(self, _exc_type=None, _exc_value=None, _exc_traceback=None):
...
         # zipkin_spans)
         if self.logging_context:
-            self.logging_context.stop()
-            self.logging_context = None
-            self.get_tracer().set_transport_configured(configured=False)
-            return
+            try:
+                self.logging_context.stop()
+            except Exception:
+                self.get_tracer().clear()
```
drolando: why not do it in the finally clause?

py_zipkin/zipkin.py
```diff
@@ -494,10 +493,18 @@ def stop(self, _exc_type=None, _exc_value=None, _exc_traceback=None):
...
+                err_msg = 'Error while emitting zipkin trace. {}'.format(
+                    self.service_name,
+                )
+                log.error(err_msg)
```
drolando: This is not a very useful message as it doesn't tell you what went wrong. Can we at least print the exception `repr`?

tests/zipkin_test.py
```diff
@@ -578,8 +578,31 @@ def test_stop_with_error(self):
             })
             assert len(span_storage) == 0
 
+    def test_error_stopping_log_context(self):
+        '''Tests if exception is raised while emitting traces that
+           tracer is cleaned u
```
drolando: up

tests/zipkin_test.py
```diff
@@ -578,8 +578,31 @@ def test_stop_with_error(self):
             })
             assert len(span_storage) == 0
 
+    def test_error_stopping_log_context(self):
+        '''Tests if exception is raised while emitting traces that
```
drolando: Can we do this as part of the integration test? So that it actually resembles what a user would do?

py_zipkin/zipkin.py
```diff
@@ -494,10 +493,19 @@ def stop(self, _exc_type=None, _exc_value=None, _exc_traceback=None):
...
         # zipkin_spans)
         if self.logging_context:
-            self.logging_context.stop()
-            self.logging_context = None
-            self.get_tracer().set_transport_configured(configured=False)
-            return
+            try:
+                self.logging_context.stop()
+            except Exception as ex:
+                err_msg = 'Error emitting zipkin trace, service:{}, {}'.format(
```
drolando: nit: I wouldn't emit the service name here, it's a bit weird. We'll need to add it in yelp_pyramid anyway.

drolando:
Can you also squash your commits?

PR #124, opened 2019-04-04
Replace ERROR_KEY: error_msg to ERROR_KEY: error_msg.encode('UTF-8')

LifeIsABoat:
This PR is Replace ERROR_KEY: error_msg to ERROR_KEY: error_msg.encode('UTF-8').

When I use it in my working company,my program needs to report a error at a particular time, the Japanese and Chinese error message were garbled.

After I changed ERROR_KEY: error_msg to ERROR_KEY: error_msg.encode('UTF-8'), the Japanese and Chinese error message turned to be normal.

cc @adriancole for reviewing.Thank you.

### DIFF
--- a/tests/zipkin_test.py
+++ b/tests/zipkin_test.py
@@ -578,22 +578,6 @@ class TestZipkinSpan(object):
             })
             assert len(span_storage) == 0
 
-         with mock.patch.object(context, 'update_binary_annotations') as mock_upd:
-            context.start()
-            context.stop(ValueError, 'おやすみ')
-            assert mock_upd.call_args == mock.call({
-                zipkin.ERROR_KEY: 'ValueError: おやすみ'
-            })
-            assert len(span_storage) == 0
-
-        with mock.patch.object(context, 'update_binary_annotations') as mock_upd:
-            context.start()
-            context.stop(ValueError, '晚安')
-            assert mock_upd.call_args == mock.call({
-                zipkin.ERROR_KEY: 'ValueError: 晚安'
-            })
-            assert len(span_storage) == 0
-
     def test_error_stopping_log_context(self):
         '''Tests if exception is raised while emitting traces that
            1. tracer is cleared
--- a/tests/zipkin_test.py
+++ b/tests/zipkin_test.py
@@ -578,6 +578,22 @@ class TestZipkinSpan(object):
             })
             assert len(span_storage) == 0
 
+         with mock.patch.object(context, 'update_binary_annotations') as mock_upd:
+            context.start()
+            context.stop(ValueError, 'おやすみ')
+            assert mock_upd.call_args == mock.call({
+                zipkin.ERROR_KEY: 'ValueError: おやすみ'
+            })
+            assert len(span_storage) == 0
+
+        with mock.patch.object(context, 'update_binary_annotations') as mock_upd:
+            context.start()
+            context.stop(ValueError, '晚安')
+            assert mock_upd.call_args == mock.call({
+                zipkin.ERROR_KEY: 'ValueError: 晚安'
+            })
+            assert len(span_storage) == 0
+
     def test_error_stopping_log_context(self):
         '''Tests if exception is raised while emitting traces that
            1. tracer is cleared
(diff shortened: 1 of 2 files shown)


drolando:
Thanks for the PR!

You'll need to fix also `tests/zipkin_test.py::TestZipkinSpan::test_stop_with_error` since it's now failing.
While you're changing it, can you add some unicode chars in that test error string to make sure we don't break this again in the future?

tests/zipkin_test.py
```diff
@@ -578,6 +578,22 @@ def test_stop_with_error(self):
             })
             assert len(span_storage) == 0
 
+         with mock.patch.object(context, 'update_binary_annotations') as mock_upd:
```
drolando: I think there's an extra space in this line which breaks the indentation

drolando:
There are still some broken tests since in python3 that's no longer a binary string but a `str` now.

You can run `make test` on your laptop to make sure all tests pass. (you'll need to install tox first though)

LifeIsABoat:
Thank you.
I'll test it in two days.
But I can't test it today because I have something else to do.

adriancole:
thanks for the help!

LifeIsABoat:
When I write this, 

>             if int(sys.version[0:1]) >= 3:
>                self.update_binary_annotations({
>                    ERROR_KEY: error_msg,
>                })
>            else:
>                self.update_binary_annotations({
>                    ERROR_KEY: error_msg.encode('UTF-8'),
 >               })
the following information is prompted in the test results. Where did I write it wrong?
`ERROR: InvocationError for command /home/travis/build/Yelp/py_zipkin/.tox/py34/bin/coverage report -m --show-missing --fail-under 100 (exited with code 2)
ERROR:   py34: commands failed
The command "tox" exited with 1.`
py27 is the same error message.

drolando:
We fail the tests if coverage is less than 100%. The problem with that code is that python2 will always enter the else branch, py3 always the first branch and so they will complain that you didn't test for the other case.

What we usually do in these cases is to add a `  #pragma: no cover` comment on each branch so that the tests don't complain. Look at https://github.com/Yelp/py_zipkin/blob/master/py_zipkin/encoding/__init__.py#L33 for an example. Also, can you use `six.PY2` to check for the python version?

drolando:
I had a look into this change and as usual I'm super confused by how python2 handles unicode strings.
The current code works perfectly if the exception message raised is a unicode string, even in python2.
```
with mock.patch.object(context, 'update_binary_annotations') as mock_upd:
    context.start()
    context.stop(ValueError, u'晚安')
    assert mock_upd.call_args == mock.call({
        zipkin.ERROR_KEY: u'ValueError: 晚安'
    })
    assert len(span_storage) == 0
```

If I don't define them as unicode strings, the code crashes on `error_msg = u'{0}: {1}'.format(_exc_type.__name__, _exc_value)` with `UnicodeDecodeError: 'ascii' codec can't decode byte 0xe3 in position 0: ordinal not in range(128)`.

@LifeIsABoat what python version are you using? Also, can you check what's the actual type of the error message that you're throwing in your code?

LifeIsABoat:
@drolando 
----
I use six.PY3 to check for the python version,but the following information is prompted in the test results.Tests is pass,Tests Coverage is 100%.Where did I write it wrong?
```
Reorder python imports...................................................Failed
hookid: reorder-python-imports
Files were modified by this hook. Additional output:
Reordering imports in py_zipkin/zipkin.py
```
----
The Python version I'm using is **Python 2.7.10**.
I don't define them as unicode strings.
The error message is `UnicodeDecodeError: 'ascii' codec can't decode byte 0xe3 in position 45: ordinal not in range(128)`
Thanks for your help!

drolando:
@LifeIsABoat I am unable to reproduce your issue. If the test throws an error with a proper unicode string (as it should) then everything works fine as is.
It the error message is instead a normal byte string then it fails on `error_msg = u'{0}: {1}'.format(_exc_type.__name__, _exc_value)`, so adding the encoding where you're adding it isn't useful.

Can you make sure that the errors are proper unicode strings? Adding `from __future__ import unicode_literals` to the file where you generate the error might be enough.

LifeIsABoat:
@drolando 

I Add `from __future__ import unicode_literals` to my file,this problem was solved.
Thanks for your help!
I will close this pull requests.

PR #129, opened 2019-05-15
Fix unicode string decoding when decoding from thrift

drolando:
Using `str(value)` fails if the string is unicode with `UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-1: ordinal not in range(128)`.

### DIFF
--- a/tests/encoding/_decoders_test.py
+++ b/tests/encoding/_decoders_test.py
@@ -169,6 +169,28 @@ class TestV1ThriftDecoder(object):
         assert local_endpoint == Endpoint('test_service', '10.0.0.1', None, 8888)
         assert remote_endpoint == Endpoint('rem_service', '10.0.0.2', None, 9999)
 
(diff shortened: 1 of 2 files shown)


askedrelic:
🐑 it

adriancole:
ugh.. followed by. I will now use this sheep it thing

tests/encoding/_decoders_test.py
```diff
@@ -169,6 +169,28 @@ def test__convert_from_thrift_binary_annotations(self):
...
+            create_binary_annotation('key3', '再见', ann_type.STRING, local_host),
+        ]
+
+        tags, local_endpoint, remote_endpoint = \
+            decoder._convert_from_thrift_binary_annotations(
+                thrift_binary_annotations,
+            )
+
+        assert tags == {
+            'key1': u'再见',
```
adriancole: maybe change to https://www.compart.com/en/unicode/U+1F411

PR #134, opened 2019-08-02
Fix import of TBinaryProtocol

cattibrie:
### Problem
Currently, py_zipkin does not use cython implementation for Thrift Binary Protocol even if it is available, because it imports python implementation directly. 

### Solution
Use another import that automatically fallbacks to python implementation when cython one is unavailable.

### DIFF
--- a/py_zipkin/thrift/__init__.py
+++ b/py_zipkin/thrift/__init__.py
@@ -4,7 +4,7 @@ import socket
 import struct
 
 import thriftpy2
-from thriftpy2.protocol.binary import TBinaryProtocol
+from thriftpy2.protocol import TBinaryProtocol
 from thriftpy2.protocol.binary import write_list_begin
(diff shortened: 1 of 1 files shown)


drolando:
This seems fine, but aren't your problem and solution kind of the opposite?

"py_zipkin does not use cython implementation for Thrift Binary Protocol even if it is available", but then the solution says that "automatically fallbacks to python implementation when cython one is unavailable".

So is the issue that we're not using the cython implementation when available or that we always try to use it even when it's not available?

drolando:
Nevermind, I re-read it and understood what you meant. It should be available in a bit as  verison `0.18.4`.

Thanks!

cattibrie:
@drolando Thank you very much! Sorry for confusing description.
This change makes thrift encoding 2 - 2.5 times faster in our case.

PR #135, opened 2019-08-09
Add py_zipkin.testing helper functions.

drolando:
For now that includes only the MockTransportHandler. I got bored of
having to copy paste it in every other library when running integration
tests. It's much better to just export it from here.


adriancole:
good idea

PR #136, opened 2019-09-20
Ensure tags are strings

drolando:
span.tags is an untyped python dictionary, meaning that the values in there could be of any type.

The schema however says that they should be strings and that's causing validation failures in other parts of our infrastructure.

@adriancole is this correct?


adriancole:
yes indeedily

https://github.com/openzipkin/zipkin-api/blob/master/zipkin2-api.yaml#L317

PR #137, opened 2019-10-14
Zipkin 706 prevent intermediate services to override a sample rate when doing zipkin_attrs override

acer618:
with zipkin_attrs override the sample_rate is defaulted to 100 percent if no explicit sample_rate is provided. While creating zipkin_span though there is an option to additionally provide a sample_rate. This change throughs a ZipkinError if both zipkin_attr override and sample_rate were specified at the same time. The sample_rate override should be part of the zipkin_attr creation.

### DIFF
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -351,18 +351,7 @@ class zipkin_span(object):
         # child span.
         if self._is_local_root_span:
 
-            # If sample_rate is set, we need to (re)generate a trace context.
-            # If zipkin_attrs (trace context) were passed in as argument there are
-            # 2 possibilities:
-            # is_sampled = False --> we keep the same trace_id but re-roll the dice
-            #                        for is_sampled.
-            # is_sampled = True  --> we don't want to stop sampling halfway through
-            #                        a sampled trace, so we do nothing.
-            # If no zipkin_attrs were passed in, we generate new ones and start a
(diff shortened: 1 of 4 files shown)


tests/zipkin_test.py
```diff
@@ -215,6 +213,17 @@ def test_init_span_storage_wrong_type(self):
             ):
                 pass
 
+    def test_error_when_sample_rate_and_attrs_override(self):
+        with pytest.raises(ZipkinError):
```
drolando: catch the exception with `as e` and then assert that the message contains `sample_rate should not be set when zipkin_attrs are provided`

py_zipkin/zipkin.py
```diff
@@ -241,6 +241,10 @@ def __init__(
                 "use the timestamp and duration parameters."
             )
 
+        if self.zipkin_attrs_override and self.sample_rate is not None:
+            raise ZipkinError(
+                'sample_rate should not be set when zipkin_attrs are provided')
```
drolando: Also add a section in DEPRECATIONS.rst on how to do the resampling now that this is gone.

tests/zipkin_test.py
acer618: done

py_zipkin/zipkin.py
acer618: done

PR #138, opened 2019-10-29
Add HTTP transport example

drolando:
cc @adriancole @as009


drolando:
Moved HTTP transport from example/ to transport.py as discussed on gitter

Issue #130, opened 2019-05-23
Add an log annotation for a span


NeonToo:
Hi. I want to add a new annotation for a span in the purpose of logging, like
```
{
    ...other fields,
    "annotations": [{
         {
             "timestamp": timestamp,
             "value": "logging info"
         }
     ]
}
```

the only possible way I found was
```
span.update_binary_annotations({'timestamp': time.time(), 'value': 'logging info'})
```

but it adds fields in the "tags", like
```
{
    ...other fields,
    "tags": {
             "timestamp": timestamp,
             "value": "logging info"
     },
    "annotations": []
}
```

Is there any direct way to update the "annotations" field?

Thanks a lot.

dbishop:
I think this is solved by the merged PR #141 and can be closed.

PR #141, opened 2020-02-21
Add a zipkin_span.add_annotation() method

dbishop:
This is the analogue to update_binary_annotations() for annotations.

### DIFF
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -457,6 +457,7 @@ class zipkin_span(object):
                 max_span_batch_size=self.max_span_batch_size,
                 firehose_handler=self.firehose_handler,
                 encoding=self.encoding,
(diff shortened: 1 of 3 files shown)


dbishop:
Looks like this is what Issue #130 wanted?

py_zipkin/zipkin.py
```diff
@@ -542,6 +543,26 @@ def update_binary_annotations(self, extra_annotations):
             # the binary annotations for the logging context directly.
             self.logging_context.tags.update(extra_annotations)
 
+    def add_annotation(self, annotation_val, timestamp=None):
```
drolando: nit: I'd rename `annotation_val` to just `value`, not a big deal though

dbishop: That makes sense!  I just renamed that parameter and pushed.

Issue #98, opened 2018-08-29
implement "b3 single" header format


adriancole:
As discussed on https://github.com/openzipkin/b3-propagation/issues/21 and first implemented here: https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/B3SingleFormat.java https://github.com/openzipkin/brave/blob/master/brave/src/test/java/brave/propagation/B3SingleFormatTest.java

Let's support at least reading "b3" header from a single string, most commonly traceid-spanid-1
It would also be nice to support optionally writing this, especially in message providers or others with constrained environments.

Brave currently has a property like this, but its name could change with feedback:
```java
    /**
     * When true, only writes a single {@link B3SingleFormat b3 header} for outbound propagation.
     *
     * <p>Use this to reduce overhead. Note: normal {@link Tracing#propagation()} is used to parse
     * incoming headers. The implementation must be able to read "b3" headers.
     */
    public Builder b3SingleFormat(boolean b3SingleFormat) {
      this.b3SingleFormat = b3SingleFormat;
      return this;
}
```

drolando:
@adriancole How would you handle the transition phase when not all services have been updated to accept the new header?

The client cannot know if the server supports it or not since we don't do any protocol negotiation. Would people need to wait until all their servers have been upgraded before changing the client code?

adriancole:
yes.. server first, just like when we did 128bit trace ID. that is why it
is a bit more urgent to update middleware.. in order to start that
transition.

dual propagation downstream is possible but yeah simpler to start with
making everything understand parsing both.

Issue #73, opened 2018-03-27
Support missing or empty X-B3-Sampled


drolando:
Right now, we are ignoring the pyramid_zipkin path/route blacklist in firehose mode, which means firehose mode is emitting spans for things like /status.

We can support this more generally by not firehose tracing when X-B3-Sampled=0 and only sampling when it is 1 or absent (see https://github.com/openzipkin/b3-propagation for details on this)

This will be a little bit tricky because pyramid_zipkin and py_zipkin do not correctly implement this nuance of the b3 spec. py_zipkin's create_http_headers_for_new_span doesn't support X-B3-Sampled being absent, and pyramid_zipkin does not set zipkin_attrs.is_sampled to None when a trace decision is deferred (i.e. absent X-B3-Sampled)

PR #143, opened 2020-02-22
Add an extract_zipkin_attrs_from_headers() helper

dbishop:
It turns out that extracting B3 headers correctly is harder than it
sounds.  This helper function implements the standard as specified at:
https://github.com/openzipkin/b3-propagation

You can pass a dict or dict-like object to the new
extract_zipkin_attrs_from_headers() function, and get back a
ZipkinAttrs instance you can trust.

In some cases, a sample_rate float and a use_128bit_trace_id flag are
required, so they appear in the function signature as well.

This should take care of single-header extraction, which is part of
Issue #98.

This may help with Issue #73 (missing or empty X-B3-Sampled).

### DIFF
--- a/py_zipkin/request_helpers.py
+++ b/py_zipkin/request_helpers.py
@@ -1,0 +1,150 @@
+# -*- coding: utf-8 -*-
+import logging
+
+from py_zipkin.util import _should_sample
+from py_zipkin.zipkin import create_attrs_for_span
+from py_zipkin.zipkin import ZipkinAttrs
+
+log = logging.getLogger(__name__)
+
+
+def _parse_single_header(b3_header):
+    """
+    Parse out and return the data necessary for generating ZipkinAttrs.
+
+    Returns a dict with the following keys:
+        'trace_id':             str or None
+        'span_id':              str or None
+        'parent_span_id':       str or None
+        'sampled_str':          '0', '1', 'd', or None (defer)
+    """
+    parsed = dict.fromkeys(("trace_id", "span_id", "parent_span_id", "sampled_str"))
+
+    # b3={TraceId}-{SpanId}-{SamplingState}-{ParentSpanId}
+    #      (last 2 fields optional)
+    # OR
+    # b3={SamplingState}
+    bits = b3_header.split("-")
+
+    # Handle the lone-sampling-decision case:
+    if len(bits) == 1:
+        if bits[0] in ("0", "1", "d"):
+            parsed["sampled_str"] = bits[0]
+            return parsed
+        raise ValueError("Invalid sample-only value: %r" % bits[0])
+    if len(bits) > 4:
+        # Too many segments
+        raise ValueError("Too many segments in b3 header: %r" % b3_header)
+    parsed["trace_id"] = bits[0]
+    if not parsed["trace_id"]:
+        raise ValueError("Bad or missing TraceId")
+    parsed["span_id"] = bits[1]
+    if not parsed["span_id"]:
+        raise ValueError("Bad or missing SpanId")
+    if len(bits) > 3:
+        parsed["parent_span_id"] = bits[3]
+        if not parsed["parent_span_id"]:
+            raise ValueError("Got empty ParentSpanId")
+    if len(bits) > 2:
+        # Empty-string means "missing" which means "Defer"
+        if bits[2]:
+            parsed["sampled_str"] = bits[2]
+            if parsed["sampled_str"] not in ("0", "1", "d"):
+                raise ValueError("Bad SampledState: %r" % parsed["sampled_str"])
+    return parsed
+
+
+def _parse_multi_header(headers):
+    """
+    Parse out and return the data necessary for generating ZipkinAttrs.
+
+    Returns a dict with the following keys:
+        'trace_id':             str or None
+        'span_id':              str or None
+        'parent_span_id':       str or None
+        'sampled_str':          '0', '1', 'd', or None (defer)
+    """
+    parsed = {
+        "trace_id": headers.get("X-B3-TraceId", None),
+        "span_id": headers.get("X-B3-SpanId", None),
+        "parent_span_id": headers.get("X-B3-ParentSpanId", None),
+        "sampled_str": headers.get("X-B3-Sampled", None),
+    }
+    # Normalize X-B3-Flags and X-B3-Sampled to None, '0', '1', or 'd'
+    if headers.get("X-B3-Flags") == "1":
+        parsed["sampled_str"] = "d"
+    if parsed["sampled_str"] == "true":
+        parsed["sampled_str"] = "1"
+    elif parsed["sampled_str"] == "false":
+        parsed["sampled_str"] = "0"
+    if parsed["sampled_str"] not in (None, "1", "0", "d"):
+        raise ValueError("Got invalid X-B3-Sampled: %s" % parsed["sampled_str"])
+    for k in ("trace_id", "span_id", "parent_span_id"):
+        if parsed[k] == "":
+            raise ValueError("Got empty-string %r" % k)
+    if parsed["trace_id"] and not parsed["span_id"]:
+        raise ValueError("Got X-B3-TraceId but not X-B3-SpanId")
+    elif parsed["span_id"] and not parsed["trace_id"]:
+        raise ValueError("Got X-B3-SpanId but not X-B3-TraceId")
+
+    # Handle the common case of no headers at all
+    if not parsed["trace_id"] and not parsed["sampled_str"]:
+        raise ValueError()  # won't trigger a log message
+
+    return parsed
+
+
+def extract_zipkin_attrs_from_headers(
+    headers, sample_rate=100.0, use_128bit_trace_id=False
+):
+    """
+    Implements extraction of B3 headers per:
+        https://github.com/openzipkin/b3-propagation
+
+    The input headers can be any dict-like container that supports "in"
+    membership test and a .get() method that accepts a default value.
+
+    Returns a ZipkinAttrs instance or None
+    """
+    try:
+        if "b3" in headers:
+            parsed = _parse_single_header(headers["b3"])
+        else:
+            parsed = _parse_multi_header(headers)
+    except ValueError as e:
+        if str(e):
+            log.warning(e)
+        return None
+
+    # Handle the lone-sampling-decision case:
+    if not parsed["trace_id"]:
+        if parsed["sampled_str"] in ("1", "d"):
+            sample_rate = 100.0
+        else:
+            sample_rate = 0.0
+        attrs = create_attrs_for_span(
+            sample_rate=sample_rate,
+            use_128bit_trace_id=use_128bit_trace_id,
+            flags="1" if parsed["sampled_str"] == "d" else "0",
+        )
+        return attrs
+
+    # Handle any sampling decision, including if it was deferred
+    if parsed["sampled_str"]:
+        # We have 1==Accept, 0==Deny, d==Debug
+        if parsed["sampled_str"] in ("1", "d"):
+            is_sampled = True
+        else:
+            is_sampled = False
+    else:
+        # sample flag missing; means "Defer" and we're responsible for
+        # rolling fresh dice
+        is_sampled = _should_sample(sample_rate)
+
+    return ZipkinAttrs(
+        parsed["trace_id"],
+        parsed["span_id"],
+        parsed["parent_span_id"],
+        "1" if parsed["sampled_str"] == "d" else "0",
+        is_sampled,
+    )
--- a/py_zipkin/util.py
+++ b/py_zipkin/util.py
@@ -58,3 +58,11 @@ def signed_int_to_unsigned_hex(signed_int):
     if hex_string.endswith("L"):
         return hex_string[:-1]
     return hex_string
+
+
+def _should_sample(sample_rate):
+    if sample_rate == 0.0:
+        return False  # save a die roll
+    elif sample_rate == 100.0:
+        return True  # ditto
+    return (random.random() * 100) < sample_rate
--- a/py_zipkin/zipkin.py
+++ b/py_zipkin/zipkin.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 import functools
 import logging
-import random
 import time
 from collections import namedtuple
 
@@ -13,6 +12,7 @@ from py_zipkin.encoding._helpers import Span
 from py_zipkin.exception import ZipkinError
 from py_zipkin.logging_helper import ZipkinLoggingContext
 from py_zipkin.storage import get_default_tracer
+from py_zipkin.util import _should_sample
 from py_zipkin.util import generate_random_128bit_string
 from py_zipkin.util import generate_random_64bit_string
 
@@ -668,7 +668,11 @@ class zipkin_server_span(zipkin_span):
 
 
 def create_attrs_for_span(
-    sample_rate=100.0, trace_id=None, span_id=None, use_128bit_trace_id=False,
+    sample_rate=100.0,
+    trace_id=None,
+    span_id=None,
+    use_128bit_trace_id=False,
+    flags=None,
 ):
     """Creates a set of zipkin attributes for a span.
 
@@ -691,16 +695,13 @@ def create_attrs_for_span(
             trace_id = generate_random_64bit_string()
     if span_id is None:
         span_id = generate_random_64bit_string()
-    if sample_rate == 0.0:
-        is_sampled = False
-    else:
-        is_sampled = (random.random() * 100) < sample_rate
+    is_sampled = _should_sample(sample_rate)
 
     return ZipkinAttrs(
         trace_id=trace_id,
         span_id=span_id,
         parent_span_id=None,
-        flags="0",
+        flags=flags or "0",
         is_sampled=is_sampled,
     )
 
--- a/tests/request_helpers_test.py
+++ b/tests/request_helpers_test.py
@@ -1,0 +1,389 @@
+# -*- coding: utf-8 -*-
+import mock
+import pytest
+
+from py_zipkin import request_helpers
+from py_zipkin.request_helpers import ZipkinAttrs
+
+
+def test_extract_zipkin_attrs_from_headers_single_sample_only():
+    attrs = request_helpers.extract_zipkin_attrs_from_headers(
+        {"b3": "0"}, use_128bit_trace_id=False
+    )
+    assert 16 == len(attrs.trace_id)
+    assert not attrs.is_sampled
+
+    attrs = request_helpers.extract_zipkin_attrs_from_headers(
+        {"b3": "0"}, use_128bit_trace_id=True
+    )
+    assert 32 == len(attrs.trace_id)
+    assert not attrs.is_sampled
+
+    # I _think_ from reading https://github.com/openzipkin/b3-propagation that
+    # sending only a "yes" sampling decision is legit and should, I guess,
+    # result in a fresh trace with a sample_rate pegged to 1.0
+    attrs = request_helpers.extract_zipkin_attrs_from_headers(
+        {"b3": "1"}, use_128bit_trace_id=False
+    )
+    assert 16 == len(attrs.trace_id)
+    assert attrs.is_sampled
+
+    attrs = request_helpers.extract_zipkin_attrs_from_headers(
+        {"b3": "1"}, use_128bit_trace_id=True
+    )
+    assert 32 == len(attrs.trace_id)
+    assert attrs.is_sampled
+
+    # Gibberish gets a None
+    assert None is request_helpers.extract_zipkin_attrs_from_headers({"b3": "yakka!"})
+
+
+@pytest.mark.parametrize(
+    "no_sample_string,yes_sample_string", [("0", "1"), ("false", "true")]
+)
+def test_extract_zipkin_attrs_from_headers_multi_sample_only(
+    no_sample_string, yes_sample_string
+):
+    # While you shouldn't encode X-B3-Sampled as true or false, a lenient
+    # implementation may accept them
+    attrs = request_helpers.extract_zipkin_attrs_from_headers(
+        {"X-B3-Sampled": no_sample_string}, use_128bit_trace_id=False
+    )
+    assert 16 == len(attrs.trace_id)
+    assert not attrs.is_sampled
+
+    attrs = request_helpers.extract_zipkin_attrs_from_headers(
+        {"X-B3-Sampled": no_sample_string}, use_128bit_trace_id=True
+    )
+    assert 32 == len(attrs.trace_id)
+    assert not attrs.is_sampled
+
+    # A specified debug flag should probably override any sample header,
+    # though this is a bit pathological of the sender to send us.
+    attrs = request_helpers.extract_zipkin_attrs_from_headers(
+        {"X-B3-Sampled": no_sample_string, "X-B3-Flags": "1"},
+        use_128bit_trace_id=False,
+    )
+    assert 16 == len(attrs.trace_id)
+    assert attrs.is_sampled
+
+    # I _think_ from reading https://github.com/openzipkin/b3-propagation that
+    # sending only a "yes" sampling decision is legit and should, I guess,
+    # result in a fresh trace with a sample_rate pegged to 1.0
+    attrs = request_helpers.extract_zipkin_attrs_from_headers(
+        {"X-B3-Sampled": yes_sample_string}, use_128bit_trace_id=False
+    )
+    assert 16 == len(attrs.trace_id)
+    assert attrs.is_sampled
+
+    attrs = request_helpers.extract_zipkin_attrs_from_headers(
+        {"X-B3-Sampled": yes_sample_string}, use_128bit_trace_id=True
+    )
+    assert 32 == len(attrs.trace_id)
+    assert attrs.is_sampled
+
+    attrs = request_helpers.extract_zipkin_attrs_from_headers(
+        {"X-B3-Flags": "1"}, use_128bit_trace_id=False
+    )
+    assert 16 == len(attrs.trace_id)
+    assert attrs.is_sampled
+
+    attrs = request_helpers.extract_zipkin_attrs_from_headers(
+        {"X-B3-Flags": "1"}, use_128bit_trace_id=True
+    )
+    assert 32 == len(attrs.trace_id)
+    assert attrs.is_sampled
+
+    # Gibberish gets a None
+    assert None is request_helpers.extract_zipkin_attrs_from_headers(
+        {"X-B3-Sampled": "yakka!"}
+    )
+
+
+def test_extract_zipkin_attrs_from_headers_invalids():
+    span_id = "7e5991634df0c66e"
+    parent_span_id = "987d98239475e0a8"
+
+    for invalid in [
+        "",
+        "a",
+        "-".join(("a", "")),
+        "-".join(("", "b")),
+        "-".join(("a", "", "whee")),
+        "-".join(("a", span_id, "whee")),
+        "-".join(("a", span_id, "whee", parent_span_id)),
+        "-".join(("a", span_id, "whee", "")),
+        "-".join(("a", span_id, "1", "c", "d?!")),
+    ]:
+        print("bad b3 header: %r" % invalid)
+        assert None is request_helpers.extract_zipkin_attrs_from_headers(
+            {"b3": invalid}, sample_rate=88.2
+        ), invalid
+
+        # Convert the (bad) input b3 header into multiple-header format input
+        bits = invalid.split("-")
+        if len(bits) == 1:
+            bad_headers = {"X-B3-TraceId": bits[0]}
+        elif len(bits) == 2:
+            bad_headers = {"X-B3-TraceId": bits[0], "X-B3-SpanId": bits[1]}
+        elif len(bits) == 3:
+            bad_headers = {
+                "X-B3-TraceId": bits[0],
+                "X-B3-SpanId": bits[1],
+                "X-B3-Sampled": bits[2],
+            }
+        elif len(bits) == 4:
+            bad_headers = {
+                "X-B3-TraceId": bits[0],
+                "X-B3-SpanId": bits[1],
+                "X-B3-Sampled": bits[2],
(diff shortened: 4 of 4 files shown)


py_zipkin/zipkin.py
```diff
@@ -639,6 +640,12 @@ def __init__(self, *args, **kwargs):
         super(zipkin_server_span, self).__init__(*args, **kwargs)
 
 
+def choose_sampling(sample_rate):
```
drolando: nit: maybe call this `should_sample`? choose_sampling gives me the idea that it'll return the sample rate.

Also, I'd add a leading `_` to this function since it doesn't seem something that's useful outside of this code.

py_zipkin/zipkin.py
```diff
@@ -709,3 +713,125 @@ def create_http_headers_for_new_span(context_stack=None, tracer=None):
         'X-B3-Flags': '0',
         'X-B3-Sampled': '1' if zipkin_attrs.is_sampled else '0',
     }
+
+
+def is_valid_trace_id(candidate):
```
drolando: Please remove validation or at least make it optional and disabled by default. Even though the spec says that ids should be 16 or 32 bytes long, the zipkin UI and api accepts also shorter ids and pads them with leading 0s.

This has unfortunately led to systems in our infrastructure using shorter ids, which wouldn't pass this validation.

Your call, but I'd just get rid of the validation which will also simplify your tests a lot

py_zipkin/zipkin.py
```diff
@@ -709,3 +713,125 @@ def create_http_headers_for_new_span(context_stack=None, tracer=None):
...
+
+
+def _extract_zipkin_attrs_from_single_header(b3_header, sample_rate,
+                                             use_128bit_trace_id):
+    # b3={TraceId}-{SpanId}-{SamplingState}-{ParentSpanId}
+    # where the last two fields are optional.
+    bits = b3_header.split('-')
+    if len(bits) == 1:
+        # See if it's a valid encoding of (only) sample decision
+        if bits[0] == 'false' or bits[0] == '0':
```
drolando: I don't think `false` is a valid sample header. Reading https://github.com/openzipkin/b3-propagation/blob/master/RATIONALE.md it seems like it could be either `0`, `1` or `d` but the description of this case is very unclear.
cc @adriancole

py_zipkin/zipkin.py
```diff
@@ -709,3 +713,125 @@ def create_http_headers_for_new_span(context_stack=None, tracer=None):
...
+def is_valid_span_id(candidate):
+    return candidate and re.match(r'[0-9a-f]{16}$', candidate)
+
+
+def _extract_zipkin_attrs_from_single_header(b3_header, sample_rate,
+                                             use_128bit_trace_id):
+    # b3={TraceId}-{SpanId}-{SamplingState}-{ParentSpanId}
+    # where the last two fields are optional.
+    bits = b3_header.split('-')
+    if len(bits) == 1:
```
drolando: Add a comment that there could also be just 1 field and that's the sampling decision. Otherwise this if seems in contradiction with the above comment that says only the last 2 fields are optional

py_zipkin/zipkin.py
```diff
@@ -709,3 +713,125 @@ def create_http_headers_for_new_span(context_stack=None, tracer=None):
...
+            use_128bit_trace_id=use_128bit_trace_id)
+    if not is_valid_trace_id(bits[0]) or not is_valid_span_id(bits[1]):
+        return None
+    if len(bits) == 4:
+        if not is_valid_span_id(bits[3]):
+            return None
+        parent_span_id = bits[3]
+    else:
+        parent_span_id = None
+    if len(bits) >= 3 and bits[2]:
```
drolando: add an empty line before this

py_zipkin/zipkin.py
```diff
@@ -709,3 +713,125 @@ def create_http_headers_for_new_span(context_stack=None, tracer=None):
...
+        elif bits[2] == '0' or bits[2] == 'false':
+            is_sampled = False
+        else:
+            # Not valid according to spec
+            return None
+    else:
+        # sample flag missing; means "Defer" and we're responsible for
+        # rolling fresh dice
+        is_sampled = choose_sampling(sample_rate)
+    return ZipkinAttrs(bits[0], bits[1], parent_span_id,
```
drolando: one arg per line

py_zipkin/zipkin.py
```diff
@@ -709,3 +713,125 @@ def create_http_headers_for_new_span(context_stack=None, tracer=None):
...
+    The input headers can be any dict-like container that supports "in"
+    membership test and a .get() method that accepts a default value.
+
+    Returns a ZipkinAttrs instance or None
+    """
+    if 'b3' in headers:
+        return _extract_zipkin_attrs_from_single_header(
+            headers['b3'], sample_rate, use_128bit_trace_id)
+
+    # No single header, let's get old-school
```
drolando: since you have an helper function for `_extract_zipkin_attrs_from_single_header` I'd add one also for the regular headers

py_zipkin/zipkin.py
```diff
@@ -709,3 +713,125 @@ def create_http_headers_for_new_span(context_stack=None, tracer=None):
...
+        # We have 1==Accept, 0==Deny
+        # Also, "While you shouldn't encode X-B3-Sampled as true or false, a
+        #        lenient implementation may accept them."
+        if sampled_str == '1' or sampled_str == 'true':
+            is_sampled = True
+        elif sampled_str == '0' or sampled_str == 'false':
+            is_sampled = False
+        else:
+            # Not valid according to spec
+            return None
```
drolando: Maybe log a warning in these cases where the headers are not valid?

tests/zipkin_test.py
```diff
@@ -885,3 +885,464 @@ def test_create_headers_for_new_span_custom_tracer(gen_mock):
...
         tracer=tracer,
     )
+
+
+@mock.patch('py_zipkin.zipkin.create_attrs_for_span')
+def test_extract_zipkin_attrs_from_headers_single_sample_only(mock_create_attrs):
+    mock_create_attrs.return_value = 'xyz'  # just a sentinel
+    # While you shouldn't encode X-B3-Sampled as true or false, a lenient
+    # implementation may accept them
+    for no_sample_string in ('0', 'false'):
```
drolando: use pytest.mark.parameterize instead

tests/zipkin_test.py
```diff
@@ -885,3 +885,464 @@ def test_create_headers_for_new_span_custom_tracer(gen_mock):
     assert expected == zipkin.create_http_headers_for_new_span(
         tracer=tracer,
     )
+
+
+@mock.patch('py_zipkin.zipkin.create_attrs_for_span')
+def test_extract_zipkin_attrs_from_headers_single_sample_only(mock_create_attrs):
+    mock_create_attrs.return_value = 'xyz'  # just a sentinel
```
drolando: I'd get rid of this and let the function return a real ZipkinAttr.

In the cases where you don't care about the exact value of the ids (like when you only pass the sample flag), just check that specific value.

This will also get rid of the extra `assert mock_create_attrs.mock_calls` making the tests easier to read

py_zipkin/zipkin.py
```diff
@@ -709,3 +713,125 @@ def create_http_headers_for_new_span(context_stack=None, tracer=None):
         'X-B3-Flags': '0',
         'X-B3-Sampled': '1' if zipkin_attrs.is_sampled else '0',
     }
+
+
```
drolando: This file is getting very long and these functions don't really depend on anything else defined in here. What about creating a new file, something like request_helpers.py or so

dbishop: Will do!  I'm wrangling the structure around, then I'll start in on all the other (great!) suggestions.

py_zipkin/zipkin.py
dbishop: Makes sense that single `b3` header would not have the legacy baggage of the old `X-B3-Sample` header.

Removing that leniency.

py_zipkin/zipkin.py
dbishop: No prob, ripping it out :)

py_zipkin/zipkin.py
dbishop: Done

py_zipkin/zipkin.py
dbishop: Done

py_zipkin/zipkin.py
dbishop: Done (including additional cleanup & commenting)

py_zipkin/zipkin.py
dbishop: Done

py_zipkin/zipkin.py
dbishop: I did, but it's literally just repeating the function signature of the enclosing function.

```
    return _extract_zipkin_attrs_from_multiple_headers(
        headers, sample_rate, use_128bit_trace_id)
```

I could extract the values from headers and then call a helper function, but then it'd need to take 5 args.  Not sure that helps readability, either.

py_zipkin/zipkin.py
dbishop: Done (throughout).  I picked `log.warning(...)` let me know if another level is more appropriate.

tests/zipkin_test.py
dbishop: Done

tests/zipkin_test.py
dbishop: Done

tests/request_helpers_test.py
```diff
@@ -0,0 +1,408 @@
+# -*- coding: utf-8 -*-
+import mock
+import pytest
+
+import py_zipkin.request_helpers as request_helpers
```
drolando: why not `from py_zipkin import request_helpers`?

tests/request_helpers_test.py
```diff
@@ -0,0 +1,408 @@
...
+from py_zipkin.request_helpers import ZipkinAttrs
+
+
+def test_extract_zipkin_attrs_from_headers_single_sample_only():
+    # While you shouldn't encode X-B3-Sampled as true or false, a lenient
```
drolando: this comment is now wrong

tests/request_helpers_test.py
```diff
@@ -0,0 +1,408 @@
...
+    assert ZipkinAttrs(
+        trace_id, span_id, None, '0', False,
+    ) == request_helpers.extract_zipkin_attrs_from_headers({
+        'X-B3-TraceId': trace_id,
+        'X-B3-SpanId': span_id,
+    }, sample_rate=88.2)
+    assert [
+        mock.call()
+    ] == mock_random.mock_calls
+    mock_random.reset_mock()
```
drolando: why do you care? The response is correct and I don't think these extra asserts add much value and make the tests even longer

tests/request_helpers_test.py
```diff
@@ -0,0 +1,408 @@
...
+        assert ZipkinAttrs(
+            trace_id, span_id, None, '0', True,
+        ) == request_helpers.extract_zipkin_attrs_from_headers({
+            'X-B3-TraceId': trace_id,
+            'X-B3-SpanId': span_id,
+            'X-B3-Sampled': yes_trace_str,
+        }, sample_rate=88.2)
+        assert [] == mock_random.mock_calls
+
+    # non-defer debug-trace, no parent span
```
drolando: there's no `d` in here

tests/request_helpers_test.py
```diff
@@ -0,0 +1,408 @@
...
+        'X-B3-ParentSpanId': parent_span_id,
+    }, sample_rate=88.2)
+    assert [] == mock_random.mock_calls
+
+
+@mock.patch('py_zipkin.util.random.random')
+def test_extract_zipkin_attrs_from_headers_single(mock_random):
+    # b3={TraceId}-{SpanId}-{SamplingState}-{ParentSpanId}
+    # where the last two fields are optional.
+
```
drolando: Do we actually need all these tests to cover all the cases? There's a lot of them and they seem to repeat themselves.

I think the repeating part stems from the fat that we're basically implementing the same logic twice. Could we refactor away the common logic in request_helpers.py so that we only have to test it once? The only difference between the 2 afaict is whether all the fields are in a single header or in multiple ones.

So we could just have the logic to parse the headers t be separate and then call the same underlying function to generate the ZipkinAttrs

tests/request_helpers_test.py
dbishop: Whoops, just a brain fart; fixing

tests/request_helpers_test.py
dbishop: Fixed

tests/request_helpers_test.py
dbishop: Fair point; removing.

tests/request_helpers_test.py
dbishop: I went back and double-checked https://github.com/openzipkin/b3-propagation and the "d" thing is only for single-header AFAICT.

For multi-header, it says only (no mention of "d" for Sampling State nor Debug Flag):
```
Sampling State

An accept sampling decision is encoded as X-B3-Sampled: 1 and a deny as X-B3-Sampled: 0. Absent means defer the decision to the receiver of this header. For example, a Sampled header might look like: X-B3-Sampled: 1.

Note: Before this specification was written, some tracers propagated X-B3-Sampled as true or false as opposed to 1 or 0. While you shouldn't encode X-B3-Sampled as true or false, a lenient implementation may accept them.
Debug Flag

Debug is encoded as X-B3-Flags: 1. Debug implies an accept decision, so don't also send the X-B3-Sampled header.
```

tests/request_helpers_test.py
dbishop: The specification is a little loose and I kept running into wacky under-specified edge cases.  I think the volume of tests is justified by that.

Personally, I'd rather have overly-exhaustive tests than untested edge cases.

Because of the spec, it's not _quite_ apples to apples.  The single-header form is newer and drops some backwards-compat, the "d" thing is only in single-header, "flags" isn't a thing in single-header.

I guess those impedance mismatches could be smoothed over, but because of the spec, you'd really want the tests like they are to make sure _that_ code is right and each side of the spec (multi-header/single-header) are correct.

Refactoring the implementation to have fewer lines seems reasonable for reasons like clarity, DRY, less bug surface area, etc., but I'm not sure having shorter tests is a great justification.

Let me see what the refactoring looks like/

dbishop: @drolando That was a great idea, and I think the code cleaned up nicely.  Please see latest commit.

I still the the tests are good, as they stand now, for the reasons I mentioned earlier.

adriancole: ps only one flag value was ever defined, 1, so not sure how much under-specifying we'd need to do to clarify d is same as presence of that flag.. can we fix the spec somehow?

py_zipkin/request_helpers.py
```diff
@@ -0,0 +1,147 @@
...
+        'sample_decision_only': False,
+    }
+    if headers.get('X-B3-Flags') == '1':
+        parsed['sampled_str'] = 'd'
+    # Unlike the single-header, "absent" and "present-but-empty-string" are
+    # both representable, and the spec only mentions Absent for encoding
+    # "Defer".  So that means we need to translate an empty-string to a
+    # non-empty, but invalid string.
+    if parsed['sampled_str'] == '':
+        parsed['sampled_str'] = 'INVALID'
```
adriancole: why not use exception handling or just quit out.. this is corrupt.

py_zipkin/request_helpers.py
```diff
@@ -0,0 +1,147 @@
...
+    parsed['sample_decision_only'] = (
```
adriancole: if this is actually needed, it seems more like a utility than something needed in a data structure.. that or maybe it is conventional in python to do things like this.

py_zipkin/request_helpers.py
```diff
@@ -0,0 +1,147 @@
...
+        all(parsed[k] is None for k in ('trace_id', 'span_id', 'parent_span_id'))
+        and parsed['sampled_str'])
+    return parsed
+
+
+def extract_zipkin_attrs_from_headers(headers, sample_rate=100.0,
```
adriancole: why does extracting use a sample rate? The result of extracting headers is just the data extracted.. turning the headers into a new context sure.. but why are the concerns conflated here?

py_zipkin/request_helpers.py
```diff
@@ -0,0 +1,147 @@
...
+
+    Returns a ZipkinAttrs instance or None
+    """
+    if 'b3' in headers:
+        parsed = _parse_single_header(headers['b3'])
+    else:
+        parsed = _parse_multi_header(headers)
+
+    # Handle the lone-sampling-decision case:
+    if parsed['sample_decision_only']:
```
adriancole: traceId and spanId are required unless it is sampling decision only. you can just check for traceId after ensuring your above implementations don't accidentally allow an invalid/corrupt extraction to return

py_zipkin/request_helpers.py
```diff
@@ -0,0 +1,147 @@
...
+        # See if it's a valid encoding of (only) sample decision
+        if parsed['sampled_str'] in ('0', 'false'):
```
adriancole: it is more common to have separate keys.. one for sampled and one for debug, prioritizing debug as it implies sampled. In cases where formats permit old handling (only b3 multi would allow false), you'd keep parsing coupled to that, vs later here. That means the parse result would have a tri-state boolean for sampled. ex the 'sampled' key doesn't exist, or it does and it is true or false (or some explicit encoding of that ex 1 or 0)

py_zipkin/request_helpers.py
```diff
@@ -0,0 +1,147 @@
...
+            sample_rate = 0.0
+        elif parsed['sampled_str'] in ('1', 'true', 'd'):
```
adriancole: debug wins always, so this moves to the top.. it is also unintuitive to handle false before true..

py_zipkin/request_helpers.py
```diff
@@ -0,0 +1,147 @@
...
+            sample_rate = 100.0
+        else:
+            # Nonsense gets a None
+            log.warning('Ignoring invalid sampling decision %r; '
```
adriancole: logs like this should occur at parsing as for example, b3 single has a smaller amount of valid options than multi. Once you verify you can't return nonsense from your own impl of b3 single and multi, you'd have no reason to double check here as the implementation is sealed to this file.

py_zipkin/request_helpers.py
```diff
@@ -0,0 +1,147 @@
...
+            return None
+        attrs = create_attrs_for_span(
+            sample_rate=sample_rate,
+            use_128bit_trace_id=use_128bit_trace_id,
+            flags='1' if parsed['sampled_str'] == 'd' else '0',
+        )
+        return attrs
+
+    # Enforce requirement of having trace & span IDs
+    if not parsed['trace_id'] or not parsed['span_id']:
```
adriancole: again this should happen in the actual format extraction. this could be a function used in both places after you've handled the sampled-only branch.

py_zipkin/request_helpers.py
```diff
@@ -0,0 +1,147 @@
...
+        log.warning('Ignoring invalid or missing TraceId & SpanId:'
+                    '%r, %r', parsed['trace_id'], parsed['span_id'])
+        return None
+
+    # Handle sampling decision
+    if parsed['sampled_str']:
```
adriancole: not sure why we are doing this again.. similar comment to elsewhere.. be tight where the actual parsing occurs (ex you can log the input there if needed). then you don't need to do it again later.

py_zipkin/request_helpers.py
```diff
@@ -0,0 +1,147 @@
...
+            is_sampled = False
+        else:
+            # Not valid according to spec
+            log.warning('Invalid SamplingState %r; requires '
+                        '"0", "1", or "d".', parsed['sampled_str'])
+            return None
+    else:
+        # sample flag missing; means "Defer" and we're responsible for
+        # rolling fresh dice
+        is_sampled = _should_sample(sample_rate)
```
adriancole: I'd do this externally, especially as sampling like this is only one form of sampling. ex later it could be done via http parameters. having the sample rate code here is unnecessarily coupled.

py_zipkin/util.py
```diff
@@ -58,3 +58,11 @@ def signed_int_to_unsigned_hex(signed_int):
     if hex_string.endswith('L'):
         return hex_string[:-1]
     return hex_string
+
+
+def _should_sample(sample_rate):
```
adriancole: similar to above, this shouldn't be a part of a change around parsing..

tests/request_helpers_test.py
dbishop: (I misspoke earlier and said `"flags" isn't a thing in single-header`; I meant `"flags" isn't a thing in multi-header`)

I'm not sure I understand what you're saying.  I think the spec and the code of this PR are clear in that:
1) single-header b3 has a tri-state `SamplingState` segment with "1", "0", and "d" defined where "d" means "debug" and is like a "1" with a bit of additional semantic import.
2) multi-header has an ill-fated `X-B3-Flags` header where `1` means `d` (i.e. debug), and apparently no other flags are defined (not even `0` for "there are no flags" or "debug is not active"?)

As far as fixing the spec, I can only assume other folks have already coded to it and the cat is out of the bag?  Like the true/false thing: I decided to go ahead and be lenient in both encodings here wrt `true`/`false` values of the sampling decision; after the refactoring, either both encodings accept them or neither do.  I decided that being lenient would be better.

py_zipkin/request_helpers.py
```diff
@@ -0,0 +1,147 @@
...
+def _parse_multi_header(headers):
+    """
+    Parse out and return the data necessary for generating ZipkinAttrs.
+
+    Returns a dict with the following keys:
+        'trace_id':             str or None
+        'span_id':              str or None
+        'parent_span_id':       str or None
+        'sampled_str':          '0', '1', 'd', or None
+        'sample_decision_only': True if header contains only a sampling
```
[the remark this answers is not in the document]
dbishop: The difference is that when only a sampling decision is specified, trace_id/span_id need to be generated (the call to `create_attrs_for_span()`).

Do you mean the logic to determine if we're in that state (sample decision only) should live in `extract_zipkin_attrs_from_headers()` instead of the parser helpers?

py_zipkin/request_helpers.py
```diff
@@ -0,0 +1,147 @@
...
+        'span_id': headers.get('X-B3-SpanId', None),
+        'parent_span_id': headers.get('X-B3-ParentSpanId', None),
+        'sampled_str': headers.get('X-B3-Sampled', None),
+        'sample_decision_only': False,
+    }
+    if headers.get('X-B3-Flags') == '1':
+        parsed['sampled_str'] = 'd'
+    # Unlike the single-header, "absent" and "present-but-empty-string" are
+    # both representable, and the spec only mentions Absent for encoding
+    # "Defer".  So that means we need to translate an empty-string to a
```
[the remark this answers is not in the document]
dbishop: Just because you and I both guessed right doesn't make the spec unambiguous.  Is it necessary for specifications to be unambiguous? No, but it simplifies implementing them. I'd expect any disambiguation you could do in the spec would help anyone trying to implement it, not just me.

py_zipkin/request_helpers.py
dbishop: Oh, I wouldn't blame the language for my infelicities ;)

This is the second time you mentioned something about this key, so I'll see about moving that logic out of the parsing.

py_zipkin/request_helpers.py
dbishop: ...because I didn't think of it.  But that's a great idea!  I'll have the parsers use exceptions to bail on invalid input and that might simplify the common post-parsing logic.

py_zipkin/request_helpers.py
dbishop: Well, this is my call-site, which I'm pretty happy with:
```
# Note: this is a monkey-patch for eventlet.wsgi.HttpProtocol.handle_one_response
# Also api.sample_rate_pct is a global for the process set via config file
def _patched_handle_one_response(self):
    zipkin_attrs = api.extract_zipkin_attrs_from_headers(
        self.headers, sample_rate=api.sample_rate_pct, use_128bit_trace_id=True)
    ...
    with api.ezipkin_server_span(
        service_name=api.default_service_name(),
        span_name=self.command,
        zipkin_attrs=zipkin_attrs,
        sample_rate=None if zipkin_attrs else api.sample_rate_pct,
...
    ) as zipkin_span:
        ...
```
Is there something undesirable there that I'm missing?

py_zipkin/request_helpers.py
dbishop: Right!  This is getting back to the raise-exception-on-bad-input in the parsers things.  Love it!

py_zipkin/request_helpers.py
dbishop: I don't follow your first sentence.  But I think you're saying you don't want `true`/`false` accepted in the single-header case?

py_zipkin/request_helpers.py
dbishop: I haven't found that to be the case, myself, but it's easy enough for me to change the order.

py_zipkin/request_helpers.py
dbishop: Okay, I think this is getting back to the theme of validating the input in the parsers, including logging bad output, and using exceptions for flow control.  I dig it!

py_zipkin/request_helpers.py
dbishop: Makes sense.

py_zipkin/request_helpers.py
dbishop: Well, the `ZipkinAttrs` are a namedtuple, which is a tuple, which is immutable.  If a `ZipkinAttrs` gets out of here with a `is_sampled` field that just needs to get filled in immediately, a whole new ZipkinAttrs would need to be created.

That just seems a bit byzantine to me...

I'll refer you to my example call-site above in another comment and ask again how that could be better?  I'm pretty happy with it.

py_zipkin/request_helpers.py
dbishop: I'm sensing a theme about handing bad input in the parsing, perhaps with exceptions for flow control, simplifying the common code post-parsing...

py_zipkin/util.py
dbishop: similar to above, I think it's pretty good and don't understand the rationale for not having the `ZipkinAttrs` that come out being ready to feed into `zipkin_span()`.

dbishop:
@adriancole thanks for the prompt review, I think validating in the parsers, moving the logging there, and using exceptions for flow control will simplify things nicely!

I'm working up a new commit for that now.  👍

py_zipkin/request_helpers.py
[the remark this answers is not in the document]
adriancole: I mean it is surprising to have a field that is basically caching a result of a function you can do anytime after the parse. it being present in the data structure is distracting.

py_zipkin/request_helpers.py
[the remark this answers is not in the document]
adriancole: I think it is overly pendantic, but I'll update the spec to explicitly say invalid data is not to be interpreted the same as if you never sent the header.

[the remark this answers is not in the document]
adriancole: https://github.com/openzipkin/b3-propagation/pull/40

py_zipkin/request_helpers.py
adriancole: in other codebases, we reuse code like you've written in scenarios that don't assume a specific sampling rate implementation. To make things more flexible, we compose the feature to extract the headers with whatever it is that needs it (ex applying a rate). By keeping these things separate, we can also clarify the unit tests which have little need to concern themselves with rate.

py_zipkin/request_helpers.py
adriancole: sorry I'll be more concrete

parsed['sampled'] = true or false regardless of the shape of headers, unless the sampled value was absent.
parsed['debug'] = true only when 'd' in single format or iff 'X-B3-Flags: 1' in multi

basically the header extract function cleans the data into atoms like above.

py_zipkin/request_helpers.py
adriancole: I'll leave this point up to @drolando as it is more about maintenance than anything else, and I don't maintain this particular code.

adriancole:
thanks for the work here and good luck!

dbishop:
New commit up; I'll reiterate that based on how many edge cases got caught just in this clean-up, the level of testing here is quite appropriate.

py_zipkin/request_helpers.py
dbishop: Interesting.

Like you, I don't maintain this codebase either.  But this is a new helper-function.  In the past, anyone using `py_zipkin` had to write their own functions to do this.  So no existing users will be hampered by any lack of flexible composition here.

I think anyone needing that use-case could perform follow-on work here to add that complexity when it's necessary.

One of the things I liked about `py_zipkin` when looking for a python tracing library was how it did nearly all I wanted it to (subject to a couple open PRs, hehe), and not much more.

py_zipkin/request_helpers.py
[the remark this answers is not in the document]
dbishop: Thanks!  I think it's probably an "agree to disagree" situation on the level of pedantry here. :)

py_zipkin/request_helpers.py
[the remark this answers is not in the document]
dbishop: I was indeed able to move the necessary (subject to our disagreement about whether this code should be flexibly composed in a more general way) logic for this elsewhere and get rid of the silly dict key.

py_zipkin/request_helpers.py
dbishop: Ah, okay, makes sense.  I think the new code does this.  In both multi and single-header cases, the sampling decision is normalized to `'0', '1', 'd', or None (defer)`

adriancole: sound good

dbishop:
I think this is cleanly rebased against black-formatted `master` branch.  Please let me know if you want me to do anything else on this PR.  I think the tests, though pretty long and exhaustive, did help guide implementation changes, refactorings, etc. and have great value for catching regressions.

PR #145, opened 2020-02-23
Provide a TracingThread subclass of threading.Thread

dbishop:
If you had a span context "open" and then ran code in multiple threads,
which themselves opened spans, they would end up as time-ordered
children of each other instead of all being sibling spans whose
parentIDs were all equal to the parent thread's span context's spanID.

That didn't make sense to me, so if you use
py_zipkin.storage.TracingThread instead of threading.Thread, then the
Tracers all work together to have the right span tree.

See tests/integration/multithreading_test.py which may do a better job
explaining this than the above copy.

Also added a function, py_zipkin.zipkin.has_default_trace() to
"non-destructively" probe whether a Tracer instance has been created.
Users of py_zipkin can use this to conditionalize code, but it is also
used by TracingThread to to nothing if an existing Trace instance
doesn't already exist.  In that case, each thread will get their own
local-root-spans, which I think is the correct behavior in that case.

In service of the above, Tracer and Stack classes now both have a copy()
method that does a deep-copy of the context stack.  I don't expect that
to get used outside of py_zipkin internals themselves.

### DIFF
--- a/py_zipkin/instrumentations/python_threads.py
+++ b/py_zipkin/instrumentations/python_threads.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+import threading
+from functools import partial
+
+from py_zipkin import storage
+
+_orig_Thread_start = threading.Thread.start
+_orig_Thread_run = threading.Thread.run
+
+
+def _Thread_pre_start(self):
+    self._orig_tracer = None
+    if storage.has_default_tracer():
+        self._orig_tracer = storage.get_default_tracer().copy()
+
+
+def _Thread_wrap_run(self, actual_run_fn):
+    # This executes in the new OS thread
+    if self._orig_tracer:
+        # Inject our copied Tracer into our thread-local-storage
+        storage.set_default_tracer(self._orig_tracer)
+    try:
+        actual_run_fn()
+    finally:
+        # I think this is probably a good idea for the same reasons the
+        # parent class deletes __target, __args, and __kwargs
+        if self._orig_tracer:
+            del self._orig_tracer
+
+
+def patch_threading():  # pragma: no cover
+    """Monkey-patch threading module to work better with tracing."""
+
+    def _new_start(self):
+        _Thread_pre_start(self)
+        _orig_Thread_start(self)
+
+    def _new_run(self):
+        _Thread_wrap_run(self, partial(_orig_Thread_run, self))
+
+    threading.Thread.start = _new_start
+    threading.Thread.run = _new_run
+
+
+def unpatch_threading():  # pragma: no cover
+    threading.Thread.start = _orig_Thread_start
+    threading.Thread.run = _orig_Thread_run
--- a/py_zipkin/storage.py
+++ b/py_zipkin/storage.py
@@ -98,6 +98,17 @@ class Tracer(object):
         kwargs["_tracer"] = self
         return zipkin_span(*argv, **kwargs)
 
+    def copy(self):
+        """Return a copy of this instance, but with a deep-copied
+        _context_stack.  The use-case is for passing a copy of a Tracer into
+        a new thread context.
+        """
+        the_copy = self.__class__()
+        the_copy._is_transport_configured = self._is_transport_configured
+        the_copy._span_storage = self._span_storage
+        the_copy._context_stack = self._context_stack.copy()
+        return the_copy
+
 
 class Stack(object):
     """
@@ -128,6 +139,12 @@ class Stack(object):
     def get(self):
         if self._storage:
             return self._storage[-1]
+
+    def copy(self):
+        # Return a new Stack() instance with a deep copy of our stack contents
+        the_copy = self.__class__()
+        the_copy._storage = self._storage[:]
+        return the_copy
 
 
 class ThreadLocalStack(Stack):
@@ -171,6 +188,20 @@ def default_span_storage():
         "details on how to migrate to using Tracer."
     )
     return get_default_tracer()._span_storage
+
+
+def has_default_tracer():
+    """Is there a default tracer created already?
+
+    :returns: Is there a default tracer created already?
+    :rtype: boolean
+    """
+    try:
+        if _contextvars_tracer and _contextvars_tracer.get():
+            return True
+    except LookupError:
+        pass
+    return hasattr(_thread_local_tracer, "tracer")
 
 
 def get_default_tracer():
--- a/tests/integration/multithreading_test.py
+++ b/tests/integration/multithreading_test.py
@@ -1,13 +1,17 @@
 # -*- coding: utf-8 -*-
 import json
-from threading import Thread
+import random
+import threading
+import time
+from operator import itemgetter
+
+from six.moves import queue
 
 from py_zipkin import Encoding
-from py_zipkin.storage import get_default_tracer
+from py_zipkin.instrumentations import python_threads
 from py_zipkin.zipkin import zipkin_span
 from tests.test_helpers import MockTransportHandler
-
-tracer = get_default_tracer()
+from tests.test_helpers import TracingThread
 
 
 @zipkin_span(service_name="service1", span_name="service1_do_stuff")
@@ -23,7 +27,7 @@ def run_inside_another_thread(transport):
         need a way to return the results to the main thread.
     :type transport: MockTransportHandler
     """
-    with get_default_tracer().zipkin_span(
+    with zipkin_span(
         service_name="webapp",
         span_name="index",
         transport_handler=transport,
@@ -34,12 +38,12 @@ def run_inside_another_thread(transport):
 
 
 def test_decorator_works_in_a_new_thread():
-    """The zipkin_span decorator is instanciated in a thread and then run in
+    """The zipkin_span decorator is instantiated in a thread and then run in
     another. Let's verify that it works and that it stores the span in the
     right thread's thread-storage.
     """
     transport = MockTransportHandler()
-    thread = Thread(target=run_inside_another_thread, args=(transport,))
+    thread = threading.Thread(target=run_inside_another_thread, args=(transport,))
     thread.start()
     thread.join()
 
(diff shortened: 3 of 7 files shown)


py_zipkin/storage.py
```diff
@@ -195,3 +225,35 @@ def set_default_tracer(tracer):
...
+    available in the parent thread, but the context stack is a deep copy (i.e.
+    contexts pushed or popped in this thread will not affect the parent
+    thread's Tracer's context stack).
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(TracingThread, self).__init__(*args, **kwargs)
+        self.__orig_tracer = None
+        if has_default_tracer():
+            self.__orig_tracer = get_default_tracer().copy()
```
dbishop: It's arguable that the `.copy()` should take place in `.start()` in case the parent thread adds more spans between the time this thread is instantiated and when it's started.

In practice, that seems unlikely, but looking at this again, I think doing the `get_default_tracer().copy()` in `TracingThread.start()` would be better.

dbishop: Fixed this issue with a follow-up commit.

drolando:
Would it be possible to instead monkeypatch threading.Thread so that this will work with any new Thread, without having to explicitly use TracingThread? Something similar to https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-threading/opencensus/ext/threading/trace.py

The current CR works, but it requires developers to know that TracingThread exists and that they're supposed to use it.

dbishop:
TBH, the `TracingThread` class really just exists to provide a testable caller for the new `Tracer.copy()`, and how it'd get used "in the real world."

I agree that a monkey-patching interface is more useful.  I was thinking about adding that (in addition?), but wasn't sure it'd be desired.  Glad to know you'd find that useful.

This is what my new code in `swift_zipkin` looks like, taking advantage of the new `Tracer.copy()` method.  Those methods in the `swift_zipkin.api` module are just greenthread-local-storage-aware versions of their `py_zipkin.storage` counterparts.

```
from eventlet import greenthread

from swift_zipkin import api


__original_init__ = greenthread.GreenThread.__init__
__original_main__ = greenthread.GreenThread.main


def _patched__init(self, parent):
    # parent thread saves current TraceData from tls to self
    if api.has_default_tracer():
        self.zipkin_tracer = api.get_default_tracer().copy()

    __original_init__(self, parent)


def _patched_main(self, function, args, kwargs):
    # child thread inherits TraceData
    if hasattr(self, 'zipkin_tracer'):
        api.set_default_tracer(self.zipkin_tracer)

    try:
        __original_main__(self, function, args, kwargs)
    finally:
        if hasattr(self, 'zipkin_tracer'):
            del self.zipkin_tracer


def patch():
    greenthread.GreenThread.__init__ = _patched__init
    greenthread.GreenThread.main = _patched_main
```

dbishop:
@drolando Latest commit also adds a `py_zipkin.storage.patch_threading()` function that'll be easier for folks to use than having to switch existing code to use `TracingThread`.

py_zipkin/storage.py
```diff
@@ -195,3 +226,57 @@ def set_default_tracer(tracer):
...
+
+    def start(self):
+        _Thread_pre_start(self)
+        super(TracingThread, self).start()
+
+    def run(self):
+        _Thread_wrap_run(self, super(TracingThread, self).run)
+
+
+def patch_threading():  # pragma: no cover
```
drolando: can you add a new `instrumentations` folder and put this in a threading.py file? Thanks!

tests/integration/multithreading_test.py
```diff
@@ -50,3 +56,92 @@ def test_decorator_works_in_a_new_thread():
     assert len(spans) == 2
     assert spans[0]['name'] == 'service1_do_stuff'
     assert spans[1]['name'] == 'index'
+
+
+def _do_one_little_request(output_q, my_input):
+    print('_do_one_little_request(%s): %r' % (my_input, get_default_tracer()))
```
drolando: extra print

tests/integration/multithreading_test.py
```diff
@@ -50,3 +56,92 @@ def test_decorator_works_in_a_new_thread():
...
+    with zipkin_span(
+        service_name='_one_little_request',
+        span_name='do-the-thing',
+        binary_annotations={'input.was': my_input},
+    ) as span_ctx:
+        print('_do_one_little_request(%s): in new span; %r' % (
```
drolando: extra print

tests/integration/multithreading_test.py
```diff
@@ -50,3 +56,92 @@ def test_decorator_works_in_a_new_thread():
...
+    transport = MockTransportHandler()
+    with zipkin_span(
+        service_name='main_thread',
+        span_name='handle_one_big_request',
+        transport_handler=transport,
+        sample_rate=100.0,
+        encoding=Encoding.V2_JSON,
+        use_128bit_trace_id=True,
+    ) as span_ctx:
+        print('in root span; main thread: %r' % get_default_tracer())
```
drolando: extra print

tests/integration/multithreading_test.py
```diff
@@ -50,3 +56,92 @@ def test_decorator_works_in_a_new_thread():
...
+        expected_trace_id = span_ctx.zipkin_attrs.trace_id
+        expected_parent_id = span_ctx.zipkin_attrs.span_id
+
+        # Now do three subrequests
+        req_count = 3
+        threads = []
+        output_q = queue.Queue()
+        for thread_idx in range(req_count):
+            this_thread = thread_class(
+                target=_do_one_little_request, args=(output_q,
```
drolando: Can you rebase on the `use_black_for_code_style` branch and run black on these files? `tox -e black` should be enough to automatically update them

tests/integration/multithreading_test.py
```diff
@@ -50,3 +56,92 @@ def test_decorator_works_in_a_new_thread():
...
+    subrequest_spans = spans[1:]
+    assert len(subrequest_spans) == 3
+    assert parent_span['name'] == 'handle_one_big_request'
+    assert [s['name'] for s in subrequest_spans] == ['do-the-thing'] * 3
+    assert set(s['tags']['input.was'] for s in subrequest_spans) == {
+        'input-0', 'input-1', 'input-2'}
+    assert [s['traceId'] for s in subrequest_spans] == [expected_trace_id] * 3
+    # Perhaps most importantly, all the subrequest spans should share the same
+    # parentId, which is the main thread's span's 'id'
+    assert [s['parentId'] for s in subrequest_spans] == [expected_parent_id] * 3
```
drolando: can you just do a big for loop and put the asserts in there? I think it's more easily readable than all the list comprehensions

py_zipkin/storage.py
```diff
@@ -97,6 +98,16 @@ def zipkin_span(self, *argv, **kwargs):
         kwargs['_tracer'] = self
         return zipkin_span(*argv, **kwargs)
 
+    def copy(self):
+        # Return a copy of this instance, but with a deep-copied
+        # _context_stack.  The use-case is for passing a copy of a Tracer into
+        # a new thread context.
```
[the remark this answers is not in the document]
dbishop: yup, done

py_zipkin/storage.py
```diff
@@ -195,3 +226,57 @@ def set_default_tracer(tracer):
...
+    try:
+        actual_run_fn()
+    finally:
+        # I think this is probably a good idea for the same reasons the
+        # parent class deletes __target, __args, and __kwargs
+        if self._orig_tracer:
+            del self._orig_tracer
+
+
+class TracingThread(threading.Thread):
```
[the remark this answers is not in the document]
dbishop: In trying to remove it, now, it does make it harder to unit test `_Thread_pre_start` and `_Thread_wrap_run`, which is pretty handy.  I got code coverage into those, and integration tests cover it _actually_ working.

That combo seems somewhat powerful.

However, I guess I could add more `# pragma: no cover` and we just rely on the integration test?

Please advise.

py_zipkin/storage.py
dbishop: Yup!

tests/integration/multithreading_test.py
dbishop: yup!

py_zipkin/storage.py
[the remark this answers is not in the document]
dbishop: @drolando oh, hey, what about we just move `TracingThread` into the tests tree, where it can be a tool for getting the unittest coverage I like.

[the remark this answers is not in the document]
drolando: Yeah if people don't need to really use it directly I'd just move it to a test file.

tests/integration/multithreading_test.py
dbishop: fixed

tests/integration/multithreading_test.py
dbishop: fixed

tests/integration/multithreading_test.py
dbishop: @drolando This is great!  We switched our huge codebase over to pre-commit & black not too long ago, and we've been loving it.

I haven't mentioned it before, but I've been really impressed and pleased with how easy it's been to keep submissions high quality thanks to the infrastructure you've already got going here:  pre-commit, blocking on 100% line coverage in unit tests, etc.

Enforcing black formatting will make contributing even easier!

dbishop:
@drolando ok... I had a a couple bumps along the road, but I think this branch is cleaned up and gtg.  If you want to merge the black branch first, that's fine.  I'll clean up the conflict i'll get from that.

py_zipkin/instrumentations/python_threads.py
```diff
@@ -0,0 +1,47 @@
...
+    try:
+        actual_run_fn()
+    finally:
+        # I think this is probably a good idea for the same reasons the
+        # parent class deletes __target, __args, and __kwargs
+        if self._orig_tracer:
+            del self._orig_tracer
+
+
+def patch_threading():  # pragma: no cover
```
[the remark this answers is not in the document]
dbishop: No, though I did the moral equivalent in eventlet with greenthreads, and it's working well:

```
# monkeypatching eventlet.greenthread.GreenThread methods:

def _patched__init(self, parent):
    # parent thread saves current TraceData from tls to self
    if api.has_default_tracer():
        self.zipkin_tracer = api.get_default_tracer().copy()

    __original_init__(self, parent)


def _patched_main(self, function, args, kwargs):
    # child thread inherits TraceData
    # (this code executes in the new greenthread, and copies
    # the Tracer instance from self to this new thread's thread-local-storage:
    if hasattr(self, 'zipkin_tracer'):
        api.set_default_tracer(self.zipkin_tracer)

    try:
        __original_main__(self, function, args, kwargs)
    finally:
        if hasattr(self, 'zipkin_tracer'):
            del self.zipkin_tracer
```

The key, of course, is the new Tracer.copy() method.

For the code here, the Thread patcher, the integration test actually uses threads and failed without the `Tracer.copy()` method.  But that's the extent to which I've run it.

drolando:
All these new features should be available in `py-zipkin==0.19.0` in a few minutes. Thanks for all the pull requests!

PR #149, opened 2020-03-04
Add create http headers helper

drolando:
Followup to #142 

@dbishop


dbishop:
@drolando looks like a great way to handle the problem I had.  👍 Thanks for the heads-up!

PR #153, opened 2020-08-17
Support PRODUCER and CONSUMER spans

### LINKED ISSUE #151: One-way tracing
Hi, 
I am trying to enable tracing over rabbitmq messages. We have a few services that communicate over http and also send/receive events to rmq channels which then start further actions. The goal is to connect these requests and messages in the unified trace to better see causality behind certain actions.
My approach was to start a new span in publish_message / recieve_message methods and store/read zipkin headers from message headers. Example span when publishing: 

```
with zipkin_span( service_name='some name', span_name='some name', kind=Kind.PRODUCER ):
     headers = create_http_headers_for_new_span()

```
Receiving message is similar but there I put kind=Kind.CONSUMER.
I have a problem with kind option here is the exact error:

`ERROR: Error emitting zipkin trace. ZipkinError("(logging_helper.py:110) KeyError: <Kind.PRODUCER: 'PRODUCER'>") [py_zipkin.zipkin:500]`

don't understand why this is happening, are producer/consumer spans supported in version 0.20.0 ?

drolando:
Adds support for PRODUCER and CONSUMER spans. Also fixes local spans in the V1 format as cr/cs/sr/ss shouldn't be set for those afaict.

cc @gingercookiemage
Fixes #151


py_zipkin/encoding/_helpers.py
```diff
@@ -128,26 +122,22 @@ def build_v1_span(self):
...
-        )
-
-        if self.kind != Kind.LOCAL:
-            # If kind is not LOCAL, then we only want client or
-            # server side annotations.
-            for ann in _DROP_ANNOTATIONS_BY_KIND[self.kind]:
-                del full_annotations[ann]
-
-        # Add user-defined annotations. We write them in full_annotations
+        annotations = OrderedDict([])
```
drolando, line 125: @adriancole can you double check this is correct?

mr/ms spans have no 2nd annotation?

adriancole (commented):
as messaging spans also need a duration, it is sadly a little more complicated, though what you did won't break I think.

You could consider implementing the conversion fully, which allows round-tripping to work. It is a bit hard to do, but also hasn't changed in a while so not much maintenance later.

https://github.com/openzipkin/zipkin/blob/master/zipkin/src/main/java/zipkin2/v1/V2SpanConverter.java#L143-L168

https://github.com/openzipkin/zipkin/blob/master/zipkin/src/test/java/zipkin2/v1/SpanConverterTest.java#L584-L611

PR #156, opened 2020-11-25
Test against python3.8

dbaxa:


dbaxa:
Testing with python3.8 seems to work fine. While there are some portions of thriftpy2 that use the deprecated `@coroutine` (https://github.com/Thriftpy/thriftpy2/issues/144) the usage in this repository doesn't seem to trigger/reference the related code paths (although I could be wrong about this).

dbaxa:
cc @drolando

drolando:
@dbaxa thanks! Sorry I missed your ping

PR #157, opened 2021-03-01
[WIP]API to support storage of zipkin tags

acer618:


drolando:
This is a hack that only we need (and only in 1 service), so don't add it here. I'd rather look into whether we can add `span` as an argument to https://github.com/Yelp/pyramid_zipkin/blob/master/pyramid_zipkin/tween.py#L199 and keep this extra global variable in yelp_zipkin_utils

PR #166, opened 2022-06-08
Drop Python 2.7 support

yaroliakh:
Dropping python 2.7 which also fixes incompatibility with newer protobuf library versions


acer618 (approved):
I'd update Changelog.rst and setup.py after the merge. When i pull from master and want to add the tags.
Otherwise looks good.

PR #165, opened 2022-06-02
Rebuild protobuf using protoc 3.19

xiao-chen:
Previous attempt at https://github.com/Yelp/py_zipkin/pull/163 failed tests since unit tests doesn't check `installed` (which makes sense...).

https://github.com/Yelp/py_zipkin/pull/164/files tries to pin protobuf version to prevent running into this problem, but that limits the protobuf version.

It seems like rebuilding the protobuf file with protoc 3.20 can make this work, as suggested by the error message.
```
18:20:54 E   If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
```

PR #167, opened 2022-06-15
Remove py27 from .travis.yml

xiao-chen:
More discussions in https://github.com/Yelp/py_zipkin/pull/165.
I think this will help release 1.0.0 to pypi

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,6 @@ matrix:
     env: TOXENV=flake8
   - python: 3.7
     env: TOXENV=black
-  - python: 2.7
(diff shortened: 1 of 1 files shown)


xiao-chen:
@yaroliakh @acer618 could you please take a look? Just wanted to have 1.0.0 on https://pypi.org/project/py-zipkin/#history so I can pick this up downstream. Thanks in advance

yaroliakh:
@xiao-chen we're in process of migration from travis to GHA. Once this is done new version will be published to pypi.

PR #171, opened 2022-08-30
drop python 3.5

benbariteau:
I thought about dropping python 3.6 as well, but that would require jumping through a bunch of hoops for supporting pypy3 still.


matthewess (approved):
lg2m!

ymilki (approved):

ymilki (changes requested):
Need to drop py35 from github actions too

ymilki (approved):

PR #172, opened 2022-08-30
Add type annotations

benbariteau:
Some stuff worth noting:
- `thrift-pyi` was used to generate the type annotations for the thrift setup
- `mypy-protobuf` was used to generate the type annotations for the protobuf setup. `protoc` now supports it natively, but it's quite new and was missing some features `py-zipkin` depends on.
- fixes the broken `make install-hooks` target
- adds a development virtualenv
- adds a way to rebuild the protobuf without also updating it
- moves the updating of the protobuf into the main Makefile
- there's several places which use `Optional` due to protobuf and thrift not requiring values to be provided
- there's several places which use `Union[str, bytes]` due to allowing encoding/decoding in text or bytes
- I made `TypeGuard` functions as needed. They could maybe be better organized, but I think keeping them near where they are used makes sense.
- `thrift-pyi` doesn't deal with bytes vs. str values properly, which is an open issue. There's a few situations I have to do a type ignore because of this.
- `thrift-pyi` also doesn't deal with some enum types properly, so I had to ignore those types as well
- `thirftpy2` doesn't have type stubs, so I just ignored it in the `mypy.ini`
- I changed the protobuf installation check and logic so that I wouldn't have to liberally place assertions around for type correctness
- `py_zipkin/instrumentations/python_threads.py` is doing really weird stuff, so I kind of gave up on trying to type it
- `Stack._storage` required refactoring as the subclass used `@property` to override, which is not currently supported. I simply made the base class use a `property` descriptor instead of a normal attribute to fix the issue.
- I ended up doing a really strange pattern in `py_zipkin/thrift/__init__.py`, where I "import" zipkinCore during type checking and it's defined by thrift during execution. This has to do with the fact that `thrift-pyi` pretends it's a module, but it's actually an attribute of the module. This is transparent outside the module, but inside we have to do this weird pattern.
- Relatedly, since `thrift-pyi` names the fake module `zipkinCore` instead of `zipkin_core`, as we originally did, I changed the attribute name to match. This means changing all the references as well.


py_zipkin/encoding/_encoders.py
```diff
@@ -185,15 +245,44 @@ def _create_json_endpoint(self, endpoint, is_v1):
...
+
+class JSONv1Span(TypedDict, total=False):
+    traceId: str
+    name: Optional[str]
+    id: Optional[str]
+    annotations: List[JSONv1Annotation]
+    binaryAnnotations: List[JSONv1BinaryAnnotation]
+    parentId: str
+    timestamp: int
+    duration: int
```
matthewess, line 274: ++

py_zipkin/thrift/__init__.py
```diff
@@ -1,26 +1,45 @@
...
-zipkin_core = thriftpy2.load(thrift_filepath, module_name="zipkinCore_thrift")
+
+# this is an extremely weird pattern, but since zipkinCore isn't a "real"
+# module, but the .pyi file pretends it is, we can't import it for real but we
+# can during type checking. Hence, we end up with this weird pattern.
+if TYPE_CHECKING:  # pragma: no cover
+    from . import zipkinCore
+else:
+    # load this as `zipkinCore` so that thrift-pyi generation matches
+    zipkinCore = thriftpy2.load(thrift_filepath, module_name="zipkinCore_thrift")
```
matthewess, line 29: Should I file a ticket for circling back to think about a better way to handle this situation? I think we could write a slightly cleaner pattern:
```python
pkg/protocol/pkg.thrift
pkg/protocol/__init__.py
pkg/protocol/pkg.pyi
```
...where `__init__.py` abstracts away the `TYPE_CHECKING` check and returns either a stub module or a runtime thrift module. sort of a combination of the [pattern followed in `monk`](https://sourcegraph.yelpcorp.com/python-packages/monk/-/blob/monk/util.py#L40-52) and the pattern shown here: https://github.com/unmade/thrift-pyi#user-content-quickstart

py_zipkin/thrift/__init__.py
```diff
@@ -115,35 +149,49 @@ def annotation_list_builder(annotations, host):
...
     # TODO: Remove the type hard-coding of STRING to take it as a param option.
-    ann_type = zipkin_core.AnnotationType.STRING
+    ann_type = zipkinCore.AnnotationType.STRING
     return [
         create_binary_annotation(key, str(value), ann_type, host)
         for key, value in binary_annotations.items()
     ]
 
 
+class SpanKwargs(TypedDict, total=False):
```
matthewess, line 173: 10/10 class name

matthewess (commented):
overall lg2m! do we want to gitignore the pyi stubs? i could see an argument either way, since we are already committing some other generated files in this repo, but i don't think the stubs should be necessary for consumers of this package?

py_zipkin/thrift/__init__.py
benbariteau: Yeah, that's a good idea.

benbariteau:
> overall lg2m! do we want to gitignore the pyi stubs? i could see an argument either way, since we are already committing some other generated files in this repo, but i don't think the stubs should be necessary for consumers of this package?

My thought is that is _possible_ consumers will use the thrift or protobuf types directly so keeping them in the setup makes sense and checking them in fits with the pattern for the generated files as well.

matthewess (approved):

py_zipkin/thrift/__init__.py
matthewess: CORESERV-11991

ymilki (approved):

py_zipkin/encoding/_decoders.py
```diff
@@ -102,20 +112,29 @@ def _convert_from_thrift_endpoint(self, thrift_endpoint):
...
+        :type thrift_annotations: list of zipkinCore.Span.Annotation
+        :returns: (annotations, local_endpoint, kind, timestmap, duration)
         """
         local_endpoint = None
         kind = Kind.LOCAL
-        all_annotations = {}
-        timestamp = None
-        duration = None
+        all_annotations: Dict[str, Optional[int]] = {}
+        timestamp: Optional[int] = None
```
acer618, line 133: Optional[float] ?

acer618, line 133: Optional[float] ?

py_zipkin/encoding/__init__.py
```diff
@@ -58,7 +61,9 @@ def detect_span_version_and_encoding(message):
     raise ZipkinError("Unknown or unsupported span encoding")
 
 
-def convert_spans(spans, output_encoding, input_encoding=None):
+def convert_spans(
+    spans: bytes, output_encoding: Encoding, input_encoding: Optional[Encoding] = None
+) -> Union[str, bytes]:
```
acer618, line 66: List[Union[str, bytes]] ?

Makefile
```diff
@@ -12,7 +12,7 @@ install:
 	pip install .
 
 install-hooks:
-	tox -e pre-commit -- install -f --install-hooks
+	tox -e pre-commit
```
acer618, line 15: for my understanding - what is the reason for removing "-- install -f --install-hooks"?

py_zipkin/thrift/zipkinCore.pyi
```diff
@@ -0,0 +1,45 @@
...
+    BYTES = 1
+    I16 = 2
+    I32 = 3
+    I64 = 4
+    DOUBLE = 5
+    STRING = 6
+
+@dataclass
+class Endpoint:
+    ipv4: Optional[int] = None
```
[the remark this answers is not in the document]
benbariteau: Yeah this is autogenerated. Apparently we binary encode the IP address.

py_zipkin/encoding/_decoders.py
benbariteau: Timestamps are ints of microseconds in this context.

Makefile
benbariteau: The `pre-commit` env in the `tox.ini` passes additional positional arguments to `pre-commit run --all-files`. I think this target was written before a change in that setup when this made more sense.

py_zipkin/encoding/__init__.py
benbariteau: `encoder.encode_queue` converts a `List[Union[str, bytes]]` into `Union[str, bytes]`

py_zipkin/encoding/protobuf/__init__.py
```diff
@@ -32,7 +44,27 @@ def encode_pb_list(pb_spans):
     return pb_list.SerializeToString()
 
 
-def create_protobuf_span(span):
+class ProtobufSpanArgsDict(TypedDict, total=False):
```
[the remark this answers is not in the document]
benbariteau: `create_protobuf_span` encodes these to bytes.

acer618 (approved):

PR #176, opened 2022-11-17
fix build

benbariteau:
This
- max pins thriftpy2, as the newer versions break some tests
- max pins mypy as the newest version has new failures
- switches the flake8 repo to the github one from the now removed gitlab one


ymilki (approved):

jvperrin (approved):
Makes sense, I think I saw some errors related to this with marley where down-pinning mypy fixed it, but a more systematic fix would be great

acer618 (approved):

PR #181, opened 2023-02-03
Drop support for Thrift

dmfallak:
The latest version of Zipkin has deprecated support for Thrift encoding: https://zipkin.io/zipkin/2.11.3/zipkin/zipkin2/codec/Encoding.html .

Supporting Thrift here also imposes maintenance difficulties, as different versions of thriftpy2 are incompatible, and later versions of thriftpy2 have an issue with serializing/deserializing BinaryAnnotations. Dropping the dependency on thriftpy2 makes it easier for projects that require py_zipkin to keep it up to date.


py_zipkin/encoding/_decoders.py
```diff
@@ -1,39 +1,22 @@
...
     """Creates encoder object for the given encoding.
-
     :param encoding: desired output encoding protocol
     :type encoding: Encoding
     :return: corresponding IDecoder object
     :rtype: IDecoder
     """
     if encoding == Encoding.V1_THRIFT:
-        return _V1ThriftDecoder()
+        raise NotImplementedError(f"{encoding} decoding not yet implemented")
```
acer618, line 19: nit: should we call this "decoding no longer supported" to be consistent with the message in the encoder

py_zipkin/encoding/_encoders.py
```diff
@@ -26,7 +25,7 @@ def get_encoder(encoding: Encoding) -> "IEncoder":
     :rtype: IEncoder
     """
     if encoding == Encoding.V1_THRIFT:
-        return _V1ThriftEncoder()
+        raise NotImplementedError(f"{encoding} encoding no longer supported")
     if encoding == Encoding.V1_JSON:
         return _V1JSONEncoder()
```
acer618, line 30: Can we also drop V1 JSON Encoder be needed?

acer618 (approved):

acer618 (approved):

PR #184, opened 2023-03-23
Add back exports in py_zipkin.encoding

dmfallak:
### Problem
Some of the functions were removed because they were no longer used in the file, but turns out they are exported and used outside the file.

### Solution
Add the exported functions back to the file.

### Testing
`make && make test`


jvperrin (approved):
Nice, thanks for the fix!

